PR #6, opened 2015-07-03
run gofmt on the src

csquared:
I have my vim configured to automatically run `gofmt` on every save so I wind up running gofmt and polluting my feature/fix branch. Instead let's just run gofmt once over everything so we can use it from here on in.


vvo:
Seems legit, every go project should be gofmt, thoughts @Nagriar?

vvo:
Merging this for ease of review fo #7

PR #7, opened 2015-07-03
Search result struct

csquared:
This creates a struct for a `SearchResult` so we don't have to keep doing the parsing into `map[string]interface{}` or just `interface{}`, which should be avoided anyways. 

It really cleans up the search code!

See: https://github.com/usecanvas/canvas-cli/commit/f335b461f5739710ab6c721231b004ed5e9ba2c9

### DIFF
--- a/algoliasearch/Client.go
+++ b/algoliasearch/Client.go
@@ -1,148 +1,147 @@
 package algoliasearch
 
 import (
-"crypto/hmac"
-"crypto/sha256"
-"encoding/hex"
-"errors"
-"time"
-"reflect"
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"reflect"
+	"time"
 )
 
 type Client struct {
-  transport *Transport
+	transport *Transport
 }
 
 func NewClient(appID, apiKey string) *Client {
-  client := new(Client)
-  client.transport = NewTransport(appID, apiKey)
-  return client
+	client := new(Client)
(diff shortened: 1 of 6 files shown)


csquared:
Note: this is "stacked" on top of: https://github.com/algolia/algoliasearch-client-go/pull/6

vvo:
Seems legit, every go project should be gofmt, thoughts @Nagriar ?

vvo:
@csquared great, can you rebase on master so that we only have the relevant commit now? Thx

Nagriar:
Thanks @csquared,

I agree it's really easier to use a structure than a `interface{}` but the structure of the json can change and new attributes can be added or for exemple when you pass the query parameter `getRankingInfo=1`, you can retrieve a lot information. So I prefer to keep both the search with interface{} and the search with the structure for the common use.

You modification in Transport.go will affect all API call we should have a different handling for the `search` method and for the other calls.

csquared:
@Nagriar Great point -> I never should have made this the default return type for all requests! Whoops!

So are you proposing two endpoints, one that returns an `inteface{}` and one that returns a `struct`? 

Personally, I'd say to just figure out what the full response looks like and make a struct that handles it, leaving fields that did not get returned initialized to their empty state.  Ie: I'd still rather do `if result.Key != ""` than `if result.Key.(string)`.

Anyways, lemme know what you think about the above choices.

csquared:
Going to close and will let you know what I wind up doing w. this.

PR #8, opened 2015-07-30
simplify transport. should close response bodies.

maerlyn5:
The transport is a shared resource that can be used from many threads. It doesn't make a lot of sense to be changing its values underneath inflight requests. If different timeout values are needed the client should hold onto multiple transports.

The client of http.Do is responsible for closing the resp body regardless of the http status.

### DIFF
--- a/algoliasearch/Transport.go
+++ b/algoliasearch/Transport.go
@@ -41,7 +41,7 @@ func NewTransport(appID, apiKey string) *Transport {
 	tr := &http.Transport{
 		DisableKeepAlives:     false,
 		MaxIdleConnsPerHost:   2,
-		TLSHandshakeTimeout:   time.Second,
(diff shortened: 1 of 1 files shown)


maerlyn5:
By specifying a custom http.Tansport you have created way more trouble than value. The nil dialer has no timeout and caused our instances to rack up ~100k go routines of deadlocked dials. I would HIGHLY suggest you use the default transport and set a single timeout value on the http.Client directly. If the library really needs different timeouts for different types of requests store a couple different http.Clients in the algolia Client.

redox:
Thank you for contributing @maerlyn5 ! I'll let @Nagriar deal with the review & merge because I don't know enough the timeouts handling.

Nagriar:
Thanks @maerlyn5 for the fix and the explanation!

I'll add different client to handle the timeout.

PR #11, opened 2015-10-12
Update Quick Start section of README

zclancy:
Seems as though this section of the Quick Start guide wasn't updated when the ```Query``` method was renamed to ```Search```.


vvo:
Thanks! We use a readme generator so I used this instead, but thx for issue

PR #16, opened 2016-03-09
Read Response's Body to completion before closing it

aseure:
The Response's Body wasn't read to completion before being closed. It led to
multiple retries from the default HTTP client. This commit fixes this issue by
actually reading the Body to completion.


vvo:
Is it possible to test it?

PR #19, opened 2016-04-08
Add interfaces for Client and Index

dirkaholic:
Adding interfaces for Client and Index would increase the mocking and thereby the testing possibilities for apps using this client a lot.

### DIFF
--- a/algoliasearch/Client.go
+++ b/algoliasearch/Client.go
@@ -11,6 +11,27 @@ import (
 	"encoding/base64"
 )
 
+type AlgoliaClient interface {
+	SetExtraHeader(key string, value string)
+	SetTimeout(connectTimeout int, readTimeout int)
+	ListIndexes() (interface{}, error)
+	InitIndex(indexName string) AlgoliaIndex
+	ListKeys() (interface{}, error)
+	MoveIndex(source string, destination string) (interface{}, error)
+	CopyIndex(source string, destination string) (interface{}, error)
+	AddKey(acl, indexes []string, validity int, maxQueriesPerIPPerHour int, maxHitsPerQuery int) (interface{}, error)
+	AddKeyWithParam(params interface{}) (interface{}, error)
+	UpdateKey(key string, acl, indexes []string, validity int, maxQueriesPerIPPerHour int, maxHitsPerQuery int) (interface{}, error)
+	UpdateKeyWithParam(key string, params interface{}) (interface{}, error)
+	GetKey(key string) (interface{}, error)
+	DeleteKey(key string) (interface{}, error)
+	GetLogs(offset, length int, logType string) (interface{}, error)
+	GenerateSecuredApiKey(apiKey string, public interface{}, userToken ...string) (string, error)
+	EncodeParams(body interface{}) string
+	MultipleQueries(queries []interface{}, optionals ...string) (interface{}, error)
+	CustomBatch(queries interface{}) (interface{}, error)
+}
+
 type Client struct {
 	transport *Transport
 }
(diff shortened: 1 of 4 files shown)


dirkaholic:
The tests are failing because of ```Client_test.go:22: Need ALGOLIA_APPLICATION_ID and ALGOLIA_API_KEY```. Any idea why it happens on travis ?

Nagriar:
> The tests are failing because of Client_test.go:22: Need ALGOLIA_APPLICATION_ID and ALGOLIA_API_KEY. Any idea why it happens on travis ?

On pull request the credential aren't available for security reason.

dirkaholic:
Hi @Nagriar, what is the sense then of running the tests via travis on every PR if they are going to fail anyway? And more importantly how is a PR evaluated in this case? You're doing it manually?

Nagriar:
It tests at least the syntax but I agree it's not the best way to do it. Yes, we check manually the PR.

aseure:
@dirkaholic I agree that using interfaces would help to write tests and mock things. However, I'm not sure the Client and Index are things we want to mock here. Could you explain or provide the tests/mocks you want to include for both the Client and the Index? The PR would me more meaningful then.

On the form, naming `AlgoliaIndex`/`AlgoliaClient` the interfaces and `Index`/`Client` the structures is confusing IMHO. I would go for `Index`/`Client` for the interfaces and `IndexImpl`/`ClientImpl` for the structures. What do you think?

dirkaholic:
OK, looking forward to your review.

dirkaholic:
@aseure I would like to use GoMock with these interfaces for example (https://github.com/golang/mock). Then I could add tests to my own application (I'm actually doing it already using my fork including the interfaces) that would look like:

```
func TestEmptyTeamsUpdateArrayDoesNotTriggerAlgoliaRequest(t *testing.T) {
	ctrl := gomock.NewController(t)
	defer ctrl.Finish()

	mockedAlgoliaClient := mocks.NewMockAlgoliaClient(ctrl)
	mockedAlgoliaClient.EXPECT().InitIndex(gomock.Any()).Times(0)

	di := &container.ContainerBag{
		AlgoliaClient: mockedAlgoliaClient,
	}

	p := service.NewAlgoliaSearchProvider(di)
	p.UpdateTeams(make([]interface{}, 0))
}
```

and similarly with the Index. I could effectively test if the right function calls are done in the Algolia client without making an actual http request. On top I think the tests for the Algolia client itself could be improved by that as well. Would be happy to add another PR with an updated test set if it is appreciated.

In regards of the naming I was looking how it is done in other go packages and mostly it is done by just exposing the interface publicly and having another private struct with the same name, like:

```
type Client interface {
	WriteRequest(*Request) error
	ReadResponse() (*Response, error)
}

func NewClient(rw io.ReadWriter) Client {
	return &client{
		reader: reader{bufio.NewReaderSize(rw, readerBuffer)},
		writer: writer{Writer: rw},
	}
}

type client struct {
	reader
	writer
}

func (c *client) WriteRequest(req *Request) error {
	...
}

func (c *client) ReadResponse() (*Response, error) {
	...
}
```

What do you think ? I guess the Impl naming is more a Java thing, no ;-) ?

aseure:
@dirkaholic Thank you for the detailed answer, it's much appreciated.

To be totally transparent, we already planned to rework our Go client, probably in the upcoming weeks. Interfaces are definitely missing here, mainly for testing purpose as you pointed out. We discussed internally about merging your PR but since it would be difficult to change it afterwards (for further retro-compatibility), we do not want to rush it and prefer to think about the exact interfaces we need first.

Other than that, I agree on lowercase naming for the internal implementation.

dirkaholic:
😿

PR #20, opened 2016-04-14
expose http.Client for modification

maerlyn5:
different use cases require different http.Clients and transports. Rather than having to proxy every possible configuration change on the client/transport, the client should be directly exposed. Also gofmt changes.

### DIFF
--- a/algoliasearch/Client.go
+++ b/algoliasearch/Client.go
@@ -3,12 +3,13 @@ package algoliasearch
 import (
 	"crypto/hmac"
 	"crypto/sha256"
+	"encoding/base64"
 	"encoding/hex"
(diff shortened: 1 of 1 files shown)


ercolanelli-leo:
Hi @maerlyn5,

Unfortunately we won't merge your PR for 2 main reasons:

- We are currently working on a full rework of the Go client, and we are willing to code a lot more against interfaces, meaning easier testing and mocking

- If ever you want to mock our client I advise you to mock directly the `Index` structure. There is little sense for you to mock the replies of the Algolia servers, but a lot more to mock what the `Index` returns

We are still glad you took over your time to submit a PR!

PR #21, opened 2016-04-22
Transport: refactor for httpClient creator + remove reflect

odeke-em:
- Centralized code for creating a new http client.
- Removed reflect calls in favor of simpler plain type switches.
- Go fmt'd code in /algoliasearch.

### DIFF
--- a/algoliasearch/Transport.go
+++ b/algoliasearch/Transport.go
@@ -12,7 +12,6 @@ import "encoding/json"
 import "strconv"
 import "errors"
 import "time"
-import "reflect"
 import "strings"
 import "fmt"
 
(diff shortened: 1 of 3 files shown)


aseure:
Hello @odeke-em. Thank you for this PR. Much appreciated. We planned to start a rework of our Go client on Tuesday. You're right in time. :)

Your PR LGTM and the tests are OK on my side (sorry but we still have to run them by hand on our side currently). However, I would like to suggest you few modifications before merging it. Please check my following notes.

algoliasearch/Transport.go
```diff
@@ -100,13 +96,14 @@ func (t *Transport) EncodeParams(params interface{}) string {
 	v := url.Values{}
 	if params != nil {
 		for key, value := range params.(map[string]interface{}) {
-			if reflect.TypeOf(value).Name() == "string" {
+			switch value.(type) {
```
aseure: Here you can simply put `switch value := value.(type)`, so that you won't need to cast `value` in the case statements. More details here: https://golang.org/doc/effective_go.html#type_switch

odeke-em: Nice. Thank you for the catch. I'll push this up in a second.

odeke-em:
@aseure thank you for the comment. PR updated PTAL.

odeke-em:
Thank you @aseure for the review and @ercolanelli-leo for the merge.

PR #27, opened 2016-04-29
Refactor `Index` and `IndexIterator` code.

aseure:


aseure:
Tests are OK. Those are only small code refactorings and Go stylization. I would like to do more but as discussed with @ercolanelli-leo, we prefer to create a new version of the client to allow us some breaking changes.

algoliasearch/index.go
```diff
@@ -102,20 +98,25 @@ func (i *Index) WaitTask(task interface{}) (interface{}, error) {
 // `timeToWait` parameter controls the first duration, in ms, to use between
 // each retry (it will be exponentiated up to 10s).
 func (i *Index) WaitTaskWithInit(taskID float64, timeToWait float64) (interface{}, error) {
-	for true {
-		status, err := i.getStatus(taskID)
-		if err != nil {
+	var err error
+	var status interface{}
```
ercolanelli-leo: Is declaring variables on the outer block a go coding style guide line?

algoliasearch/index.go
```diff
@@ -199,61 +199,60 @@ func (i *Index) DeleteByQuery(query string, params map[string]interface{}) (inte
...
 	array := objects.([]interface{})
-	queries := make([]map[string]interface{}, len(array))
-	for i := range array {
-		queries[i] = make(map[string]interface{})
-		queries[i]["action"] = actions[i]
-		queries[i]["body"] = array[i]
+	size := len(array)
+	queries := make([]map[string]interface{}, size)
+
+	for i := 0; i < size; i++ {
```
ercolanelli-leo: Why changing from the range-based loop?

ercolanelli-leo:
LGTM (beside my two questions) :+1:

algoliasearch/index.go
aseure: More like a personal taste but no, I've never seen any guideline about it.

algoliasearch/index.go
aseure: Don't know. :) Useless change, I'll put the `range` back.

PR #32, opened 2016-05-04
Type most of the client.go file

ercolanelli-leo:


ercolanelli-leo:
The way errors are handled are suited after all.

PR #33, opened 2016-05-04
Type client.go

ercolanelli-leo:


algoliasearch/client.go
```diff
@@ -48,195 +46,171 @@ func (c *Client) SetTimeout(connectTimeout int, readTimeout int) {
 
 // ListIndexes returns the list of all indexes belonging to this Algolia
 // application.
-func (c *Client) ListIndexes() (interface{}, error) {
-	return c.transport.request("GET", "/1/indexes", nil, read)
+func (c *Client) ListIndexes() (ListIndexesRes, error) {
```
aseure: You can definitely use named return values here like `(res ListIndexesRes, err error)`. That's what I do on my PR #34. It avoids to declare them in the function and to use them all along the function. I won't put this comment on every single function of the `Client` but you get the idea. Maybe just add a commit on top of this one if you think it's worth it.

aseure:
- [ ] Change  `func (c *Client) SetTimeout(connectTimeout int, readTimeout int)` to `func (c *Client) SetTimeout(connectTimeout, readTimeout int)`

algoliasearch/client.go
```diff
@@ -48,195 +46,171 @@ func (c *Client) SetTimeout(connectTimeout int, readTimeout int) {
...
+	indexes := ListIndexesRes{}
+	err = json.Unmarshal(res, &indexes)
+
+	return indexes, err
 }
 
 // InitIndex returns an Index object targeting `indexName`.
-func (c *Client) InitIndex(indexName string) *Index {
-	return NewIndex(indexName, c)
+func (c *Client) InitIndex(indexName string) Index {
```
aseure: Maybe rename `indexName` to just `name`.

algoliasearch/client.go
```diff
@@ -48,195 +46,171 @@ func (c *Client) SetTimeout(connectTimeout int, readTimeout int) {
...
-func (c *Client) ListKeys() (interface{}, error) {
-	return c.transport.request("GET", "/1/keys", nil, read)
+func (c *Client) ListKeys() (ListKeysRes, error) {
+	keys := ListKeysRes{}
+	err := c.request(&keys, "GET", "/1/keys", nil, read)
+	return keys, err
 }
 
 // MoveIndex renames the index named `source` as `destination`.
 func (c *Client) MoveIndex(source string, destination string) (interface{}, error) {
```
aseure: No need to change the type of the `interface{}` return value, I've created it in #34 so I'll change it once I'll be rebased on your branch.

algoliasearch/client.go
```diff
@@ -48,195 +46,171 @@ func (c *Client) SetTimeout(connectTimeout int, readTimeout int) {
...
-	return c.InitIndex(source).Move(destination)
+	index := c.InitIndex(source)
+	return index.Move(destination)
 }
 
 // CopyIndex duplicates the index named `source` as `destination`.
 func (c *Client) CopyIndex(source string, destination string) (interface{}, error) {
```
aseure: No need to change the type of the `interface{}` return value, I've created it in #34 so I'll change it once I'll be rebased on your branch.

algoliasearch/client.go
```diff
@@ -48,195 +46,171 @@ func (c *Client) SetTimeout(connectTimeout int, readTimeout int) {
...
-func (c *Client) AddKey(acl, indexes []string, validity int, maxQueriesPerIPPerHour int, maxHitsPerQuery int) (interface{}, error) {
-	body := map[string]interface{}{
-		"acl":                    acl,
-		"maxHitsPerQuery":        maxHitsPerQuery,
-		"maxQueriesPerIPPerHour": maxQueriesPerIPPerHour,
-		"validity":               validity,
-		"indexes":                indexes,
+// AddKey creates a new API key from the supplied `ACL` and the specified
+// optional parameters.
+func (c *Client) AddKey(ACL []string, params map[string]interface{}) (AddKeyRes, error) {
```
aseure: I don't understand this signature. Couldn't you just require from the user that the ACL are already in the `params` map? So that you won't need to add it by yourself in your function afterwards.

algoliasearch/type_query.go
```diff
@@ -0,0 +1,60 @@
+package algoliasearch
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+type MultipleQueriesRes struct {
+	Results []MultipleQueryRes `json:"results"`
```
aseure: We can discuss about that but on my side, I've chosen to directly return the slice instead of this `*Res` type. Because in the end, what the user really wants are the underlying results.

algoliasearch/client.go
ercolanelli-leo: Is this comment a mistake? Same as the previous on and I don't really see the point here.

algoliasearch/client.go
ercolanelli-leo: The thing is, with the ACL in the map it can't be really mandatory. As a parameter it definitely is.

algoliasearch/type_query.go
ercolanelli-leo: From what I remember we had this discussion already and you said that staying as close to the API as possible was better, this is why I did that. However I quite agree that this structure is useless.

algoliasearch/client.go
ercolanelli-leo: If it is to be changed anyway, I guess letting it like that is fine too, no?

algoliasearch/type_query.go
aseure: After today discussion, we agreed on only using this as an internal representation. The final result returned to the user should be the `[]MultipleQueryRes`, not the `MultipleQueriesRes`.

algoliasearch/types_query.go
```diff
@@ -0,0 +1,60 @@
...
+type QueryRes struct {
+	Hits             []map[string]interface{} `json:"hits"`
+	Page             int64                    `json:"page"`
+	NbHits           int64                    `json:"nbHits"`
+	NbPages          int64                    `json:"nbPages"`
+	HitsPerPage      int64                    `json:"hitsPerPage"`
+	ProcessingTimeMS int64                    `json:"processingTimeMS"`
+	Query            string                   `json:"query"`
+	ParsedQuery      string                   `json:"parsedQuery"`
+	Params           string                   `json:"params"`
```
aseure: `:sort` ? :)

algoliasearch/types_query.go
```diff
@@ -0,0 +1,60 @@
...
+	"strings"
+)
+
+type multipleQueriesRes struct {
+	results []MultipleQueryRes
+}
+
+type MultipleQueryRes struct {
+	QueryRes
+	Index string `json:"index"`
```
aseure: :sort ? :)

algoliasearch/types_batch.go
```diff
@@ -0,0 +1,12 @@
...
+
+type CustomBatchRes struct {
+	TaskID    int64    `json:"taskID"`
+	ObjectIDs []string `json:"objectIDs"`
+}
+
+type BatchRecord struct {
+	IndexName string                 `json:"indexName"`
+	Action    string                 `json:"action"`
+	Body      map[string]interface{} `json:"body"`
```
aseure: `:sort` ? :)

algoliasearch/types_batch.go
```diff
@@ -0,0 +1,12 @@
+package algoliasearch
+
+type CustomBatchRes struct {
+	TaskID    int64    `json:"taskID"`
+	ObjectIDs []string `json:"objectIDs"`
```
aseure: `:sort` ? :)

algoliasearch/client.go
```diff
@@ -21,222 +19,187 @@ type Client struct {
...
 }
 
 // GetLogs retrieves the `length` latest logs, starting at `offset`. Logs can
 // be filtered by type via `logType` being either "query", "build" or "error".
-func (c *Client) GetLogs(offset, length int, logType string) (interface{}, error) {
-	body := map[string]interface{}{
-		"offset": offset,
-		"length": length,
-		"type":   logType,
+func (c *Client) GetLogs(params map[string]interface{}) (res GetLogsRes, err error) {
```
aseure: Don't you prefer the original version? IMHO it was better for the end user, don't you think?

ercolanelli-leo: According to the [documentation](https://www.algolia.com/doc/rest#logs-api) all parameters are optional, this is why I chose the map.

aseure: Right.

Small improvement: what do you think about letting the user pass `nil` as a parameter for those kind of optional parameters in all the functions that require a `params map`?

ercolanelli-leo: I am not against it.

ercolanelli-leo: After thoughts maybe I am against it. That would require taking pointers, so it would need the user to declare its values before the call. Don't know if it's that easier to use.

aseure: Right, always the same "issue". Let's keep it like that then.

algoliasearch/client.go
```diff
@@ -21,222 +19,187 @@ type Client struct {
...
-	if len(userToken) == 1 {
-		userTokenStr = userToken[0]
-	} else {
-		userTokenStr = ""
+	req := make(map[string]interface{})
+	for k, v := range params {
+		req[k] = v
 	}
+	req["tagFilters"] = strings.Join(params["tagFilters"].([]string), ",")
+	message := c.transport.EncodeParams(req)
```
aseure: Why don't you simply use the original `params` map and change the `tagFilters` field?

algoliasearch/check_query.go
```diff
@@ -0,0 +1,83 @@
+
```
aseure: Missing `package algoliasearch`

algoliasearch/types_logs.go
```diff
@@ -0,0 +1,15 @@
+type GetLogsRes struct {
```
aseure: Missing `package algoliasearch`

aseure:
There is at least 2 missing `package algoliasearch`. Could you also make sure this commit compiles before we merge it.

algoliasearch/client.go
aseure: Up

PR #38, opened 2016-05-11
Fix `Index.DeleteByQuery` by using `Browse` instead of `Search`

aseure:


algoliasearch/index.go
```diff
@@ -432,16 +432,26 @@ func (i *Index) DeleteByQuery(query string, params Params) (res BatchRes, err er
 	copy["attributesToRetrieve"] = []string{"objectID"}
 	copy["hitsPerPage"] = 1000
 	copy["distinct"] = false
+	copy["query"] = query
 
-	// Find all the records that match the query
-	if _, err = i.Search(query, copy); err != nil {
+	// Retrieve the iterator to browse the results
+	if it, err = i.BrowseAll(copy); err != nil {
```
ercolanelli-leo: `it` will be scoped within the `if` so it can't be used after.

PR #40, opened 2016-05-11
tests: Adapt tests to the new client

ercolanelli-leo:


algoliasearch/client_test.go
```diff
@@ -36,852 +34,837 @@ func initTest(t *testing.T) (*Client, *Index) {
...
-	_, err = index.WaitTask(resp)
-	if err != nil {
-		t.Fatalf(err.Error())
-	}
-	items, err := index.Browse(1, 1)
+	object := Object{"name": "John Snow", "objectID": "àlgol?à"}
+
+	addWait(object, index, t)
+
+	browse, err := index.Browse(map[string]interface{}{"page": 1, "hitsPerPage": 1})
```
aseure: `_, err :=` because `browse` is unused.

algoliasearch/client_test.go
```diff
@@ -36,852 +34,837 @@ func initTest(t *testing.T) (*Client, *Index) {
...
-		t.Fatalf(err.Error())
+
+	newKey := Key{
+		ACL:                    []string{"search"},
+		Validity:               300,
+		MaxQueriesPerIPPerHour: 100,
+		MaxHitsPerQuery:        100,
 	}
 
 	acl := []string{"search"}
```
aseure: You can remove this line because `acl` is unused.

algoliasearch/index.go
```diff
@@ -445,7 +445,6 @@ func (i *Index) DeleteByQuery(query string, params Params) (res BatchRes, err er
 	copy := duplicateMap(params)
 	copy["attributesToRetrieve"] = []string{"objectID"}
 	copy["hitsPerPage"] = 1000
-	copy["distinct"] = false
```
aseure: I'm not sure we want to remove this line. Why do you want to do that?

algoliasearch/index.go
```diff
@@ -457,8 +456,8 @@ func (i *Index) DeleteByQuery(query string, params Params) (res BatchRes, err er
 	// Iterate through all the objectIDs
 	var hit map[string]interface{}
 	var objectIDs []string
-	for err != nil {
-		if hit, err = it.Next(); err != nil {
+	for err == nil {
+		if hit, err = it.Next(); err == nil {
```
aseure: Good catch! 👍

algoliasearch/index_iterator.go
```diff
@@ -23,14 +23,21 @@ func NewIndexIterator(index *Index, params map[string]interface{}) (it IndexIter
...
 		return
 	}
 
 	// If the last element of the page has been reached, the next one is loaded
-	if it.pos == it.page.NbHits {
-		if err = it.loadNextPage(); err != nil {
+	if it.pos == len(it.page.Hits) {
+		if it.cursor == "" {
+			err = errors.New("No more hits")
+
```
aseure: Extra empty line.

algoliasearch/index_iterator.go
```diff
@@ -23,14 +23,21 @@ func NewIndexIterator(index *Index, params map[string]interface{}) (it IndexIter
...
 	// Abort if the user call `Next()` on a IndexIterator that has been
 	// initialized without being able to load the first page.
-	if it.page.NbHits == 0 {
-		err = fmt.Errorf("No more hits")
+	if len(it.page.Hits) == 0 {
+		err = errors.New("No more hits")
 		return
 	}
 
 	// If the last element of the page has been reached, the next one is loaded
```
aseure: Could you add something like `or returned an error if the last element of the last page has already been returned`.

algoliasearch/client.go
```diff
@@ -160,7 +160,7 @@ func (c *Client) GenerateSecuredAPIKey(apiKey string, params Params) (key string
 // the field used to store the index name in the queries, and the strategy used
 // to perform the multiple queries.
 // The strategy can either be "none" or "stopIfEnoughMatches".
-func (c *Client) MultipleQueries(queries []map[string]interface{}, indexField, strategy string) (res []MultipleQueryRes, err error) {
+func (c *Client) MultipleQueries(queries []Params, indexField, strategy string) (res []MultipleQueryRes, err error) {
```
aseure: I intentionnaly didn't replace those types because the paramater wasn't actually a `Params` map. Thinking again about it, don't you think it would be a good idea to rename the `Params` type to `JsonMap` for instance? It would be more generic and understandable and could be used for both `queries` or `params` parameters (sorry for the nitpicking).

algoliasearch/client_test.go
```diff
@@ -67,16 +67,22 @@ func hasString(slice []string, elt string) bool {
...
 
-func checkEqual(got, expected interface{}, message string, t *testing.T) {
+func checkEqual(got, expected interface{}, message string, index Index, t *testing.T) {
 	if got != expected {
-		t.Fatalf("%s: %v expected %v", message, got, expected)
+		fatal(index, t, "%s: %v expected %v", message, got, expected)
 	}
 }
 
+func fatal(index Index, t *testing.T, fmt string, objs ...interface{}) {
```
aseure: For all the functions that require the `*testing.T`, I would suggest to put it first in the parameter list. Looks more Go-ish IMHO.

algoliasearch/client_test.go
```diff
@@ -67,16 +67,22 @@ func hasString(slice []string, elt string) bool {
...
 }
 
-func checkNbHits(got, expected int, t *testing.T) {
-	checkEqual(got, expected, "nbHits", t)
+func checkNbHits(got, expected int, index Index, t *testing.T) {
+	checkEqual(got, expected, "nbHits", index, t)
 }
 
-func checkEqual(got, expected interface{}, message string, t *testing.T) {
+func checkEqual(got, expected interface{}, message string, index Index, t *testing.T) {
```
aseure: For all the functions that require the `*testing.T`, I would suggest to put it first in the parameter list. Looks more Go-ish IMHO.

algoliasearch/client_test.go
```diff
@@ -67,16 +67,22 @@ func hasString(slice []string, elt string) bool {
 	return false
 }
 
-func checkNbHits(got, expected int, t *testing.T) {
-	checkEqual(got, expected, "nbHits", t)
+func checkNbHits(got, expected int, index Index, t *testing.T) {
```
aseure: For all the functions that require the `*testing.T`, I would suggest to put it first in the parameter list. Looks more Go-ish IMHO.

algoliasearch/index.go
ercolanelli-leo: Because distinct is an integer not a boolean, we could put 0 but I don't really see the point either.

algoliasearch/index_iterator.go
ercolanelli-leo: I am not sure to understand.

algoliasearch/client.go
ercolanelli-leo: Ok

algoliasearch/index.go
aseure: Ok for the `s/bool/integer/`. However, we must set it to 0 anyway because we don't want the user to set it to 1 in the original `params` map. That way, we ensure to override it.

algoliasearch/index_iterator.go
aseure: I was talking about the comment.

ercolanelli-leo: The main goal of this message is to differentiate a real error from the end of iteration, having a too long message isn't a good idea in my opinion

aseure: I know, I introduced it. :) I was talking about the *comment* (line 31) not the error message.

aseure:
On your last commit message: `s/index/indexes/`

aseure:
https://github.com/algolia/algoliasearch-client-go/pull/40/commits/821092684ceb01105e688b860598f0ea8cff1da2 LGTM 👍

aseure:
@redox The doc states that it's a string with 4 possible values: `min`, `strict`, `true`, `false`. Do you concur or do you mean that it can be either a string with `min` and `strict` as possible values or a boolean (not a string)?

algoliasearch/client.go
```diff
@@ -160,29 +161,25 @@ func (c *Client) GenerateSecuredAPIKey(apiKey string, params Map) (key string, e
...
 // The strategy can either be "none" or "stopIfEnoughMatches".
-func (c *Client) MultipleQueries(queries []Map, indexField, strategy string) (res []MultipleQueryRes, err error) {
-	if indexField == "" {
-		indexField = "indexName"
-	}
-
+func (c *Client) MultipleQueries(queries []IndexedQuery, strategy string) (res []MultipleQueryRes, err error) {
 	if strategy == "" {
 		strategy = "none"
+	} else if strategy != "none" && strategy != "stopIfEnoughMatches" {
```
ercolanelli-leo: For the sake of possibly hidden/debug values I wouldn't do that check

redox:
> @redox The doc states that it's a string with 4 possible values: min, strict, true, false. Do you concur or do you mean that it can be either a string with min and strict as possible values or a boolean (not a string)?

Yes.

PR #47, opened 2016-05-16
Expose interfaces instead of struct types

argon:
This allows easier mocking of Algolia within application tests

### DIFF
--- a/algoliasearch/algoliasearch.go
+++ b/algoliasearch/algoliasearch.go
@@ -1,0 +1,225 @@
+package algoliasearch
+
+// Client is a representation of an Algolia application. Once initialized it
+// allows manipulations over the indexes of the application as well as
+// network related parameters.
+type Client interface {
+
+	// SetExtraHeader allows to set custom headers while reaching out to
+	// Algolia servers.
+	SetExtraHeader(key, value string)
+
+	// SetTimeout specifies timeouts to use with the HTTP connection.
+	SetTimeout(connectTimeout, readTimeout int)
+
+	// ListIndexes returns the list of all indexes belonging to this Algolia
+	// application.
+	ListIndexes() (indexes []IndexRes, err error)
+
+	// InitIndex returns an Index object targeting `indexName`.
+	InitIndex(name string) Index
+
+	// ListKeys returns all the API keys available for this Algolia application.
+	ListKeys() (keys []Key, err error)
+
+	// MoveIndex renames the index named `source` as `destination`.
+	MoveIndex(source, destination string) (UpdateTaskRes, error)
+
+	// CopyIndex duplicates the index named `source` as `destination`.
+	CopyIndex(source, destination string) (UpdateTaskRes, error)
+
+	// AddKey creates a new API key from the supplied `ACL` and the specified
+	// optional parameters.
+	AddKey(ACL []string, params Map) (res AddKeyRes, err error)
+
+	// UpdateKey updates the API key named `key` with the supplied
+	// parameters.
+	UpdateKey(key string, params Map) (res UpdateKeyRes, err error)
+
+	// GetKey returns the ACL and validity of the API key named `key`.
+	GetKey(key string) (res Key, err error)
+
+	// DeleteKey deletes the API key named `key`.
+	DeleteKey(key string) (res DeleteRes, err error)
+
+	// GetLogs retrieves the `length` latest logs, starting at `offset`. Logs can
+	// be filtered by type via `logType` being either "query", "build" or "error".
+	GetLogs(params Map) (logs []LogRes, err error)
+
+	// GenerateSecuredAPIKey generates a public API key intended to restrict access
+	// to certain records.
+	// This new key is built upon the existing key named `apiKey`. Tag filters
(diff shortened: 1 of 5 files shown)


argon:
Tests failing as they require #40

aseure:
Hello @argon. Thank you for the great work. I'm currently working on the second version of the client with @ercolanelli-leo and we effectively planned to turn the client and the index into interfaces so your PR arrived just at the right time.

Could you just rebase your branch onto `v2` right now, fix what needs to be fixed and push-force your commit? Because we just merged #40 which brings the tests back to life and changes quite a few things (the `Index.AddKey` for instance). We'll review your PR after that and eventually merge it. :)

algoliasearch/index_iterator.go
```diff
@@ -15,21 +13,17 @@ type IndexIterator struct {
...
-func newIndexIterator(index *Index, params Map) (it IndexIterator, err error) {
-	it = IndexIterator{
+func newIndexIterator(index Index, params Map) (it IndexIterator, err error) {
+	it = &indexIterator{
 		index:  index,
 		params: duplicateMap(params),
 		pos:    0,
 	}
-	err = it.loadNextPage()
+	err = it.(*indexIterator).loadNextPage()
```
argon: I'm not entirely happy with this, but without avoiding the implicit return variable I don't see a better way.

argon:
I have included a second commit which removes the duplicated documentation from the implementation files `client.go`, `index.go`, `index_iterator.go`. You may wish to leave them in but I don't think it's wise to have the duplication and the documentation should definitely live alongside the interfaces.

algoliasearch/algoliasearch.go
```diff
@@ -0,0 +1,225 @@
+package algoliasearch
+
+// Client is a representation of an Algolia application. Once initialized it
+// allows manipulations over the indexes of the application as well as
+// network related parameters.
+type Client interface {
+
```
aseure: Extra empty line can be removed.

algoliasearch/algoliasearch.go
```diff
@@ -0,0 +1,225 @@
...
+	// DeleteByQuery finds all the records that match the `query`, according to the
+	// given 'params` and deletes them.
+	DeleteByQuery(query string, params Map) (res BatchRes, err error)
+}
+
+// IndexIterator is used by the BrowseAll functions to iterate over all the
+// records of an index (or a subset according to what the query was).
+type IndexIterator interface {
+
```
aseure: Extra empty line can be removed.

algoliasearch/index_iterator.go
aseure: Agree.

However, are you sure it's useful to have an `interface` for the `indexIterator`? In the end, it's just an implementation detail of the `index` structure. I think we should keep the `IndexIterator` as it was previously (not an interface). Final users who would like to mock the `index` will probably change all the `Browse*` functions according to their needs, without reusing the `indexIterator`. @ercolanelli-leo What do you think?

ercolanelli-leo: Since indexIterator now refers to index through its interface, maybe it's not necessary to make it itself an interface.

aseure:
@argon Thanks for the rebasing. Few things before we merge, could you:

 - Remove the two extra empty lines I pointed out
 - Revert the changes you made to the `IndexIterator` because we do not think we should use it as an interface

Other than that, everything is LGTM and the tests are OK on my side. Thank you. 👍

argon:
I strongly disagree about IndexIterator. While you're correct that it only uses the Index interface, it would result in a significant overhead when trying mock it out.

Instead of simply implementing BrowseAll and passing back a MockIndexIterator which would merely need to implement Next. Instead it's necessary to mock out Index based on the *internal* behaviour of IndexIterator when that is not something the application using the library should ever care about.

I will come up with a better solution than the force type assertion. Dropping the IndexIterator interface would be a big mistake.

aseure:
@argon My bad, you're right about the `IndexIterator` as an interface.

About the type assertion, I could suggest you to instantiate a new `*indexIterator` variable on which you can call the `loadNextPage` function without type casting and assign it to the `it` variable which is typed as a `IndexIterator` interface.

```diff
diff --git a/algoliasearch/index_iterator.go b/algoliasearch/index_iterator.go
index f6500c7..e85e993 100644
--- a/algoliasearch/index_iterator.go
+++ b/algoliasearch/index_iterator.go
@@ -14,12 +14,13 @@ type indexIterator struct {
 // to the given `params`. It is also trying to load the first page of results
 // and return an error if something goes wrong.
 func newIndexIterator(index Index, params Map) (it IndexIterator, err error) {
-       it = &indexIterator{
+       i := &indexIterator{
                index:  index,
                params: duplicateMap(params),
                pos:    0,
        }
-       err = it.(*indexIterator).loadNextPage()
+       err = i.loadNextPage()
+       it = i
        return
 }
```

argon:
As `newIndexIterator` is an internal function I have adjusted it to return an `*indexIterator` which makes sense and solves the problem.

aseure:
Tests are running OK on my side. I'll merge it now. Thank you again @argon. 👍

PR #52, opened 2016-05-23
Add test to check DNS timeout

aseure:


aseure:
From #50

algoliasearch/client_test.go
```diff
@@ -976,3 +976,28 @@ func TestSynonyms(t *testing.T) {
...
+		appID + "-2.algolianet.com",
+		appID + "-3.algolianet.com",
+	}
+
+	c := NewClientWithHosts(appID, apiKey, hosts)
+
+	start := time.Now()
+	_, _ = c.ListIndexes()
+
+	if start.Add(15 * time.Second).Before(time.Now()) {
```
ercolanelli-leo: Shouldn't we also set the timeout for the client to 15s?

aseure: It's already the case, isn't it? https://github.com/algolia/algoliasearch-client-go/blob/v2/algoliasearch/transport.go#L46

ercolanelli-leo: Ok works for me :)

ercolanelli-leo:
LGTM :+1:

Issue #68, opened 2016-06-01
Missing optional attributes in the QueryRes


Nagriar:
Here is the list of missing fields:

- offset/length (integer to specify page/hitsPerPage differently)
- message (string used to return warnings)
- facets (JObject used for the facet count)
- facets_stats (JObject same for the stats)
- exhaustiveFacetsCount (boolean to indicate if the facetsCount and facetsStats are exact)
- queryAfterRemoval (string to indicate which words are removed with removeWordsIfNoResults)
- aroundLatLng (string resolved location from the ip)
- automaticRadius (string radius computed by the engine automatically)
- index (string to indicate the index name in case of multiqueries)
- serverUsed (string the hostname of the server)
- parsedQuery (string the query after the preprocessing of the engine like normalization)
- timeoutCounts (boolean to indicate if the engine returned an exact nbHits)
- timeoutHits (boolean to indicate if the query triggered the timeout)

PR #73, opened 2016-06-01
Add missing fields to QueryRes + Fix one field (params)

aseure:
From #68


ercolanelli-leo:
I don't think there is a problem, but you confirm that default values for structure fields aren't an issue here?

aseure:
@ercolanelli-leo I'd say yes since we're talking about a response type, not a query. The two maps will just be `nil` if the response doesn't have them.

ercolanelli-leo:
Nice I just wanted to make sure I wasn't missing something

Issue #64, opened 2016-06-01
The DeleteByQuery method is wrong


Nagriar:
We can't browse the index to take all the maching objectID because if the content of the index changes between two pages, the browse response won't be exact.

So we need to wait the batch of delete to be sure we didn't miss some records and rebrowse the index until the first page of the browse is empty.

PR #76, opened 2016-06-01
Fix `Index.DeleteByQuery`

aseure:
From #64


aseure:
Migration guide updated.

Issue #62, opened 2016-06-01
Use the same naming than in the other api client for the api key methods


Nagriar:
In the other api clients the naming of the functions are: (add|update|delete)UserKey and in go we use (add|update|delete)Key. Do we want to use the v2 to also normalize the naming?

aseure:
@Nagriar As you said, normalisation is better. Could you just tell me if it only applies to the `Index` methods, not to the `Client` ones?

Nagriar:
Both at the index and client level.

aseure:
Ok on it. :)

PR #77, opened 2016-06-01
Rename API key-related methods to be more consistent with the other API clients

aseure:
From #62


aseure:
Migration wiki page updated.

PR #86, opened 2016-06-15
Allow for a custom HTTP client to be set

kujenga:
This added functionality enables the use of alternate transport mechanisms, such as those required for use within Google App Engine [1]. My company has had a branch for a while maintaining this change: https://github.com/rastech/algoliasearch-client-go/commit/59d17d0313aae8d66ed345b4737ee507d358d21e which brings in the appengine libraries as a dependency and is thus probably undesirable for the main repo. Without that [handling](https://github.com/rastech/algoliasearch-client-go/commit/59d17d0313aae8d66ed345b4737ee507d358d21e#diff-606c6ee02e002516ad4f7614a535f7e8R93) for the `*urlfetch.Transport` type, the timeouts won't be properly set, but I put a note in the method comments to that effect.

This also corrects a few minor things from golint.

[1] https://cloud.google.com/appengine/docs/go/outbound-requests

### DIFF
--- a/algoliasearch/algoliasearch.go
+++ b/algoliasearch/algoliasearch.go
@@ -1,5 +1,9 @@
 package algoliasearch
 
+import (
+	"net/http"
+)
+
 // Client is a representation of an Algolia application. Once initialized it
 // allows manipulations over the indexes of the application as well as network
 // related parameters.
@@ -10,6 +14,11 @@ type Client interface {
 
 	// SetTimeout specifies timeouts to use with the HTTP connection.
 	SetTimeout(connectTimeout, readTimeout int)
+
+	// SetHTTPClient allows a custom HTTP client to be specified.
+	// NOTE: using this may prevent timeouts set on this client from
+	// working if the underlying transport is not of type *http.Transport.
+	SetHTTPClient(client *http.Client)
 
 	// ListIndexes returns the list of all indexes belonging to this Algolia
 	// application.
--- a/algoliasearch/transport.go
+++ b/algoliasearch/transport.go
@@ -12,6 +12,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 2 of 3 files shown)


aseure:
Thank you for your PR @kujenga. We're discussing about it internally and we'll come back to you ASAP.

algoliasearch/algoliasearch.go
```diff
@@ -11,6 +15,11 @@ type Client interface {
 	// SetTimeout specifies timeouts to use with the HTTP connection.
 	SetTimeout(connectTimeout, readTimeout int)
 
+	// SetHTTPClient allows a custom http client to be specified.
```
aseure: s/http client/HTTP client/

algoliasearch/transport.go
```diff
@@ -95,8 +95,11 @@ func NewTransportWithHosts(appID, apiKey string, hosts []string) *Transport {
 
 // setTimeout changes the timeouts used by the underlying HTTP client.
 func (t *Transport) setTimeout(connectTimeout, readTimeout time.Duration) {
-	t.httpClient.Transport.(*http.Transport).TLSHandshakeTimeout = connectTimeout
-	t.httpClient.Transport.(*http.Transport).ResponseHeaderTimeout = readTimeout
+	switch transport := t.httpClient.Transport.(type) {
+	case *http.Transport:
+		transport.TLSHandshakeTimeout = connectTimeout
+		transport.ResponseHeaderTimeout = readTimeout
+	}
```
aseure: Add a `default` case displaying an error message (`Cannot use setTimeout for the underlying Transport` for instance).

aseure:
@kujenga Ok. Besides the two comments I left you, we're ready to merge. The tests are running OK.

algoliasearch/transport.go
kujenga: Seems reasonable, but how would you like me to log out the error message? I could import the stdlib [log package](https://golang.org/pkg/log/) but it doesn't look like this repo currently outputs anything, and I think adding an error return is undesired here.

aseure: We don't want to change the signature so indeed, no need to return an error here.

As you've seen, we aren't using the `log` package. I think we can simply `fmt.Fprintln(os.Stderr, ...)` here.

kujenga: Sounds good! Though I will note that you aren't currently using the `"os"` package either.

PR #87, opened 2016-06-16
Export listIndexesRes.Items so the JSON response can unmarshal into it

AndyHaskell:
This bug fix exports the `Items` field in the `listIndexesRes` struct. This makes it so calls to `ListIndexes()` will be able to get back the Algolia indexes since a field needs to be exported to be unmarshalled from JSON.


aseure:
@AndyHaskell Good catch. Thank you. I'll add a test afterwards (see #88).

Running the tests and merging your PR right after. ;)

PR #89, opened 2016-06-16
Add test for Client.ListIndexes

aseure:
From #88


aseure:
Failing because the Algolia application which runs the tests has other indexes. I need to change Travis secret credentials to have a dedicated application for the Go client.

PR #129, opened 2016-09-21
Add "noCreate" versions of PartialUpdateObject and PartialUpdateObjects

AndyHaskell:
This PR adds methods to the Algolia Go client for sending partial updates with the `createIfNotExists=false` option on individual `PartialUpdateObject` calls and the `partialUpdateObjectNoCreate` option on batch `PartialUpdateObjects` calls


aseure:
Good catch. Thank you for reporting that. We are discussing internally about how to add it to the API clients' README.

aseure:
Everything is OK on my side, except that we shouldn't have missed it in the first place. :)

Thanks a lot for reporting that @AndyHaskell. Merging it.

PR #136, opened 2016-09-28
Add GetObjectsAttrs method to get selected attributes of multiple objects

AndyHaskell:
This PR adds a variation of `GetObjects` that takes in an `attrs` parameter to allow users to retrieve only the selected attributes of multiple objects.


aseure:
Awesome. Thank you for your contributions @AndyHaskell :) LGTM 👍

PR #140, opened 2016-10-06
checkQuery: simplify ignore query conditions, reduce algorithmic complexity

odeke-em:
* Refactor checkQuery to take out the continue
label Outer, thus simplifying the code and easing readability.
* Build a map of keys to ignore upfront, so that when iterating over
queries, instead of a quadratic O(n^2) ie O(n*m) lookup time,
it becomes a constant time lookup per query so linear time O(n)
processing altogether where n = max(len(queries), len(ignore)).

### DIFF
--- a/algoliasearch/check_query.go
+++ b/algoliasearch/check_query.go
@@ -1,13 +1,20 @@
 package algoliasearch
 
+func buildKeysMap(keys []string) map[string]struct{} {
+	m := make(map[string]struct{})
+	for _, key := range keys {
+		m[key] = struct{}{}
+	}
+
+	return m
+}
+
 func checkQuery(query Map, ignore ...string) error {
-Outer:
+	ignoreKeysMap := buildKeysMap(ignore)
+
 	for k, v := range query {
-		// Continue if `k` is to be ignored.
-		for _, s := range ignore {
-			if s == k {
-				continue Outer
-			}
+		if _, shouldIgnore := ignoreKeysMap[k]; shouldIgnore {
+			continue
 		}
 
 		switch k {


aseure:
Hello @odeke-em. Thank you for your contribution!

We've reviewed your PR with @ercolanelli-leo and agreed that on a complexity point of view, your solution is better. However, `checkQuery` is an internal which is not really used a lot with the `ignore` parameters. Actually, only `GenerateSecuredAPIKey` is using it with two parameters: `"userToken"` and `"tagFilters"`, so the gain may be not that important anyway in practice.

I've benchmarked the two solutions with only two `ignore` parameters:
 - **Quadratic version** (without the map): **77.7 ns/op** over 20.000.000 loops
 - **Linear version** (with the map): **412 ns/op** over 5.000.000 loops

We are welcoming PR, but if they are intended to change non-user-facing, we would very much like some kind of metric showing an objective improvement :)

odeke-em:
Hello there @aseure, thank you for the review!

Not merging it is alright, but I just thought I'd improve the readability and also take out the label. I added the map because I noticed the amount of variables you were switching on was a lot, so figured we'd need uniqueness, but since only two values will be ignored, well and good.

Am a fan of Algolia search and in the next few days/weeks I'll be using your libraries to index content. Thanks!

aseure:
I understand @odeke-em.

Thank you for the kind words. If you need anything during your implementation, don't hesitate to reach us at support@algolia.com.

Have a nice day.

odeke-em:
Thank you @aseure, I'll do so. Have a nice day too!

Issue #142, opened 2016-10-17
Testing flags


cochran-at-niche:
When you build a Go project that uses `github.com/algolia/algoliasearch-client-go/algoliasearch` as well as the `flag` library from the standard lib, test flags (from the standard `testing` library) appear in the usage description that shows up when you provide the `-h` or `--help` flags:

```
  -test.bench string
        regular expression to select benchmarks to run
  -test.benchmem
        print memory allocations for benchmarks
  -test.benchtime duration
        approximate run time for each benchmark (default 1s)
  -test.blockprofile string
        write a goroutine blocking profile to the named file after execution
  -test.blockprofilerate int
        if >= 0, calls runtime.SetBlockProfileRate() (default 1)
  -test.count n
        run tests and benchmarks n times (default 1)
  -test.coverprofile string
        write a coverage profile to the named file after execution
  -test.cpu string
        comma-separated list of number of CPUs to use for each test
  -test.cpuprofile string
        write a cpu profile to the named file during execution
  -test.memprofile string
        write a memory profile to the named file after execution
  -test.memprofilerate int
        if >=0, sets runtime.MemProfileRate
  -test.outputdir string
        directory in which to write profiles
  -test.parallel int
        maximum test parallelism (default 4)
  -test.run string
        regular expression to select tests and examples to run
  -test.short
        run smaller test suite to save time
  -test.timeout duration
        if positive, sets an aggregate time limit for all tests
  -test.trace string
        write an execution trace to the named file after execution
  -test.v
        verbose: print additional output
```

This is due to the existence of the "testing.go" file (https://github.com/algolia/algoliasearch-client-go/blob/master/algoliasearch/testing.go), which does not follow the Go convention for test files, which specifies that all test files should end with `_test.go` (see: https://golang.org/pkg/testing/#pkg-overview). Because it does not follow the convention, it is not excluded from the build like other test files. This results in the test flags showing up in the usage description (the test flags come from the standard `testing` library, which is imported in the `testing.go` file).

Running the following minimal program with the `-h` or `--help` flags demonstrates the issue:

```
package main

import (
	"flag"

	_ "github.com/algolia/algoliasearch-client-go/algoliasearch"
)

func main() {
	flag.Parse()
}
```

Deleting the `testing.go` file fixes the issue, as does renaming it to `testing_test.go`. I suggest renaming it to follow the `_test.go` convention.

aseure:
[...]

PR #143, opened 2016-10-17
test: Rename the testing.go file into testing_test.go

aseure:
This change has been introduced to avoid including the `testing.go`
content in the build of other programs using the `algoliasearch`
package.

For instance, it was producing a side-effect with the `-h` flag when
used with the `flag` package.


aseure:
Closes #142

PR #165, opened 2016-12-02
feat/stateful-conn-with-timer

aseure:
This PR implements the new retry strategy discussed internally. It also improves the overall code of `algoliasearch/transport.go` which were the only remaining file still unmodified after the v1-v2 migration.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,20 @@
 # Changelog
 
+## 2.6.0 (2016-12-03)
+
+### Fixes
+
+- Improve the transport layer code
+- Keep the last active connection between requests to better handle DNS timeouts
+
+## 2.5.0 (2016-11-29)
+
+### Additions
+
+- Add `Index.SearchForFacetValues` method
+  + Same as `Index.SearchFacet`
+  + `Index.SearchFacet` is kept for backward-compatibility
+
 ## 2.4.0 (2016-11-02)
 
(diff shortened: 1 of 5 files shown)


aseure:
@ercolanelli-leo As this commit refactors nearly everything in `transport.go` (the only file we barely modified during the v1-v2 transition) your review would be very welcome.

@adam Could you tell me if the retry strategy looks good to you?

It can wait next week, I won't release it this week-end anyway.

aseure:
If you have time guys @ercolanelli-leo and @asurak, I would really appreciate your feedbacks on that PR.

@ElPicador you may also be interested but you already have all the Java/Scala ones to handle.

asurak:
the connection strategy looks good... you're keeping the last working host and before making a new request, you ensure this one is the first in the list of hosts, so it will be tried as first.... did I get it right?

aseure:
@asurak Yes indeed. However, as discussed with @ElPicador, I don't have a handle on the connect per se. I'm just building the list of hosts to try and perform the API call against each one. Is it enough to you or should I move to @clement-leprovost's version?

asurak:
Will the underlying runtime keep the connection open and not reconnect the next time you reuse the host?

aseure:
@asurak Yes it does.

asurak:
Then it should be fine.

Issue #287, opened 2017-01-25
AWS Lambda: socket: too many open files


sorenbs:
I am consuming events from a kinesis queue in a lambda function and calling

```go
algoliaIndex := algoliaClient.InitIndex(lambdaEvent.IndexName)
_, algoliaOperationError = algoliaIndex.AddObjects(objectsToSync(lambdaEvent))
```

After 6,930 successful calls I get the following error:

```
Cannot perform request [POST] /1/indexes/SyncItem/batch (8OVNLPT6JS-1.algolianet.com): Post https://8OVNLPT6JS-1.algolianet.com/1/indexes/SyncItem/batch: dial tcp: lookup 8OVNLPT6JS-1.algolianet.com on 10.15.0.2:53: dial udp 10.15.0.2:53: socket: too many open files
```

Is it possible that the Algolia go library doesn't correctly close old connections or am I missing something?

sorenbs:
Comment from the https://github.com/eawsy/aws-lambda-go maintainer

```
@sorenbs AWS Lambda is not freshly initialized each time you invoke it. One Lambda can persist over the time (nearly 8 hours) before being recycled. So when I see your error, it seems that you orthe algolia lib have forgotten something like defer foo.Close() somewhere. Thus, over the time the number of opened sockets grows and the error occur. But without more info, this is just an intuition.
```

aseure:
Hello @sorenbs. Sorry to hear you're experiencing this issue.

Could you tell me what is the Go client version/commit hash you are currently using in your lambda instance?

fsenart:
Voila :)

sorenbs:
@fsenart - you are amazing :-)

@aseure - I am using latest master

sorenbs:
@aseure - how quickly are you usually able to merge prs? Is it reasonable to expect it could happen today, or should I rather go ahead and use the branch for now?

aseure:
@sorenbs Merged and version bumped to `2.7.1` on `master`.

@fsenart Thank you again for the fix.

aseure:
@sorenbs Could you confirm me that the problem is fixed on your side? I'll wait for that to close your issue.

sorenbs:
It looks like this did indeed solve the issue. Thanks a lot both!

aseure:
Good to know and sorry for the inconvenience it may have caused.

sorenbs:
No problem.

This might be unrelated, but I am seeing a new network error:

```
[...]

PR #288, opened 2017-01-25
Fix unlimited socket opening problem

fsenart:
transport.go#L171 tries to make a request to each host, until one
succeed and ignore errors unless there is no host left. But
transport.go#L249 can return with an error before closing
the response at L251. This error is not caught by Transport.request()
method unless there no other host to try.
Consequence of transport.go#L249 is that with a lot number of calls we
can arrive at a situation with a lot of socket remained open.
This commit fix the problem by defering the connection closure and move
the statement before any error can happen so that in any situation the
socket is closed.

CAT: #fix
REF: algolia/algoliasearch-client-go#287


aseure:
Agree that the `Close` should have been called and deferred before the error. Good catch @fsenart and thank you for the contribution.

PR #308, opened 2017-02-24
Create *APIKey alternatives to *UserKey functions

ercolanelli-leo:
@aseure would love your feedback on this. I did not add test because in the end the old tests still work (it's a simple forwarding of the argument).


ercolanelli-leo:
@aseure done \o/

PR #309, opened 2017-03-15
Fix query format in GetObject and GetObjectsAttrs

AndyHaskell:
This PR fixes a bug retrieving attributes in the `index`'s `GetObjectsAttrs` and `GetObject` methods. In `GetObjectsAttrs`, the requested list of attributes is URL-escaped, while in `GetObject`, the attribute query params being passed into `encodeParams` are a `[]byte`, rather than a `string`. The result, in both cases, is that no attributes except the object ID are retrieved.

### DIFF
--- a/algoliasearch/index.go
+++ b/algoliasearch/index.go
@@ -45,7 +45,7 @@ func (i *index) GetObject(objectID string, attributes []string) (object Object, 
 			return
 		}
 		params = Map{
-			"attributes": attrBytes,
+			"attributes": string(attrBytes),
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


aseure:
Hello @AndyHaskell. Please excuse me for the late answer.

I'm reviewing now review your PR and get back to you soon.

Thank you for reporting this issue.

aseure:
It's a good catch that should have been spotted earlier. Thank you again for the contribution.

Actually, even the `attributesToRetrieve` from the `getObjects` is still not conform as it should be passed as a string slice, and not a slice. However, we cannot change that yet because it triggers a bug we've just discovered in the engine, thanks to you :). We're working on the fix today and I will change the type internally once it will be deployed.

Regarding this PR, it's fine to me. I'm merging it and increasing the version number of the Go client right now. Have a good day.

PR #316, opened 2017-04-04
Add standard http.ProxyFromEnvironment to default transport

zkmoney:
This is required for applications that limit outbound http/https access to a proxy server to connect to their Algolia servers, and really use this client library at all. 

The current implementation exposes a method to overwrite the underlying `*http.Client` that powers this client lib, but then when requests are made it is overwritten almost immediately whenever [`increaseDialTimeout`](https://github.com/algolia/algoliasearch-client-go/blob/master/algoliasearch/transport.go#L311) or [`resetDialTimeout`](https://github.com/algolia/algoliasearch-client-go/blob/master/algoliasearch/transport.go#L318) are called.

### DIFF
--- a/algoliasearch/transport.go
+++ b/algoliasearch/transport.go
@@ -124,6 +124,7 @@ func defaultHttpClient() *http.Client {
 // lookup timeouts).
 func defaultTransport(dialTimeout time.Duration) *http.Transport {
 	return &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
 		Dial: (&net.Dialer{
 			KeepAlive: 180 * time.Second,
 			Timeout:   dialTimeout,


aseure:
Hello @zkmoney, thank you for contributing.

I understand your issue regarding the need to set up the transport layer with a proxy. However, I don't think it's safe to enable the `ProxyFromEnvironment` by default.

You clearly identified an issue by observing that the `increaseDialTimeout` and `resetDialTimeout` functions were overriding any previously set HTTP client. Would it work for you if we fix those two functions instead? So that you could use the `SetHTTPClient` by providing it the client you want to use with `ProxyFromEnvironment` enabled?

zkmoney:
Hey @aseure, thanks for the reply!

That makes sense to me about not blindly using the built proxy func, and yes, the main concern was that custom client being overriden. It's a tough balance as your current client optimizes for what you know about your servers expectations and timeouts, etc., so is there a way to restain your expertise there while providing a way for users to just override the proxy func? If not, or if you'd rather not increase the public API footprint (understandable), then I think fixing those methods behavior to not overwrite a custom http client sounds good. 

Thanks
Z

zkmoney:
restain = retain, by the way

aseure:
I'm planning to fix the two functions in the next 24h and indeed, I'd rather prefer not to add a function simply to override the proxy function as we don't want to add a new setter for every possible field, the `SetHTTPClient` function is available for that.

Would it be OK to you?

zkmoney:
Sounds good :+1:

aseure:
Hello @zkmoney 

Coming back to you with good news. After discussing it internally with other API client maintainers, we decided it was safe to use the `ProxyFunc` by default for the transport layer. I haven't merged your own PR for this change as it was in the middle of some other fixes regarding the retry strategy. All the changes I'm mentioning are available as of version `2.12.0`, which is already available.

Regarding the two `increaseDialTimeout` and `resetDialTimeout` functions used by the retry strategy, it's now best-effort: if the `RoundTripper` of the HTTP client is a `http.Transport` (whether or not the HTTP client was provided by the user), we apply the changes to the dial timeout. Otherwise, the retry strategy won't change the timeout. This way, we're ensuring that any HTTP client set by the user won't be overridden by the retry strategy.

However, this last part is probably not relevant to you anymore as the `ProxyFunc` is now enabled by default. In that regard, I'm closing this PR but feel free to re-open it or open an issue if there is anything else you'd like to discuss.

Thank you again for contributing. Much appreciated.

zkmoney:
That sounds good to me, thanks for the update!

PR #322, opened 2017-06-06
Fix/do not ignore params for getlogs

aseure:


algoliasearch/client_test.go
```diff
@@ -199,20 +199,20 @@ func TestLogs(t *testing.T) {
...
 
 	params := Map{
-		"length": 10,
+		"length": 2,
 		"offset": 0,
 		"type":   "all",
 	}
 
-	t.Log("TestLogs: Get the last 10 logs")
+	t.Log("TestLogs: Get the last 2 logs")
```
ElPicador: I would generate 2 queries to be sure that you have 2 logs

aseure: Right, will do 👍

aseure: Done

PR #323, opened 2017-06-26
feat: Implement query rules

aseure:


algoliasearch/algoliasearch.go
```diff
@@ -313,6 +313,44 @@ type Index interface {
...
+	// arguments are optional, using an empty string for string and
+	// RulePatternAnchoring parameters and `-1` for integer parameters. For
+	// more details regarding the parameters:
+	//
+	//   - `query`:       enable full text search within the Rule fields
+	//   - `anchoring`:   restricts the search to Rules with a specific anchoring type
+	//   - `context`:     restricts the search to rules with a specific context (exact match)
+	//   - `page`:        requested page (zero-based, defaults to zero)
+	//   - `hitsPerPage`: maximum number of hits in a page (defaults to 20)
+	SearchForRules(query string, anchoring RulePatternAnchoring, context string, page, hitsPerPage int) (SearchForRulesRes, error)
```
clement-leprovost: How are you going to extend the method when we add more search parameters? Wouldn't it be safer to use a type for the parameters?

algoliasearch/types_rule.go
```diff
@@ -0,0 +1,93 @@
...
+	TaskID    int    `json:"taskID"`
+	UpdatedAt string `json:"updatedAt"`
+}
+
+type ClearRulesRes struct {
+	TaskID    int    `json:"taskID"`
+	UpdatedAt string `json:"updatedAt"`
+}
+
+type SearchForRulesRes struct {
```
clement-leprovost: What happens if the API suddenly returns more fields in the response? Is it going to cause the existing code to crash/return an error? Or will the extraneous fields simply be ignored? I am concerned about forward compatibility (i.e. making deployed code work with new versions of the API).

algoliasearch/types_rule.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// RuleCondition is the part of an Algolia Rule which describes the condition
+// for the rule. The `Context` and `Priority` are optional, hence, they will
+// get ignored if an empty string or a 0 values are respectively set to them.
+type RuleCondition struct {
+	Anchoring RulePatternAnchoring `json:"anchoring"`
+	Pattern   string               `json:"pattern"`
+	Context   string               `json:"context,omitempty"`
+	Priority  int                  `json:"priority,omitempty"`
```
clement-leprovost: Sorry, maybe it wasn't obvious in the doc, but `priority` is actually not supported. It was left out of the scope until further notice. => Please remove it.

algoliasearch/types_rule.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// NewRuleCondition generates a RuleCondition where all the possible fields can
+// be specified.
+func NewRuleCondition(anchoring RulePatternAnchoring, pattern, context string, priority int) RuleCondition {
+	return RuleCondition{
+		Anchoring: anchoring,
+		Pattern:   pattern,
+		Context:   context,
+		Priority:  priority,
```
clement-leprovost: Same as above.

algoliasearch/algoliasearch.go
aseure: I didn't know it was subject to modifications. It's already painful to pass 5 parameters. I'll make it take a map and perform a type check on the known parameters. This way, even if a new parameter is used by the user, it will let it go through.

algoliasearch/types_rule.go
aseure: Simply ignored until we implement it in the Go client, as shown here:
https://play.golang.org/p/dSJkBWxmpw

aseure:
@clement-leprovost My last commit should fix the concerns you had. Let me know.

aseure:
@clement-leprovost 

I've squashed a bit my commits, let the consequence's params be regular search query parameters and renamed `SearchForRules`/`SearchRules` to match the other API clients naming in the documentation.

Could you take a last look at this PR before I merge it?

Thank you in advance :)

PR #325, opened 2017-07-24
Improve DeleteByQuery performance

poweroftrue:
As discussed here: https://github.com/algolia/algoliasearch-client-ruby/pull/211 this the same implementation for go.

Update: I've added wait for DeleteObjects to match [documentation](https://www.algolia.com/doc/api-client/go/indexing/#delete-by-query), but I think it should be like other functions and like documentation said: "All write operations in Algolia are asynchronous by design".
```go
func (i *index) DeleteByQuery(query string, params Map) (res BatchRes, err error) {
	copy := duplicateMap(params)
	copy["hitsPerPage"] = 1000
	copy["query"] = query
	copy["distinct"] = 0

	var browseRes BrowseRes
	var objectIDs []string
	var cursor string

	for {
		// Start browsing the content by cursor
		if browseRes, err = i.Browse(copy, cursor); err != nil {
			return
		}

		// Break if there's no more matching records
		if len(browseRes.Hits) == 0 {
			break
		}

		// Collect all objectIDs
		for _, hit := range browseRes.Hits {
			objectIDs = append(objectIDs, hit["objectID"].(string))
		}

		// Set the new cursor from response
		cursor = browseRes.Cursor
	}

	// Delete all the objects
	if res, err = i.DeleteObjects(objectIDs); err != nil {
		return
	}

	return
}
```

### DIFF
--- a/algoliasearch/index.go
+++ b/algoliasearch/index.go
@@ -457,6 +457,7 @@ func (i *index) DeleteByQuery(query string, params Map) (err error) {
 	copy["distinct"] = 0
 
 	var browseRes BrowseRes
+	var batchRes BatchRes
 	var objectIDs []string
 	var cursor string
 
@@ -481,8 +482,13 @@ func (i *index) DeleteByQuery(query string, params Map) (err error) {
 	}
 
 	// Delete all the objects
-	if _, err = i.DeleteObjects(objectIDs); err != nil {
+	if batchRes, err = i.DeleteObjects(objectIDs); err != nil {
 		return
+	}
+
+	// Wait until DeleteObjects completion
+	if err := i.WaitTask(batchRes.TaskID); err != nil {
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


aseure:
Hello @poweroftrue,

Thank you for your contribution. You're absolutely right, this way the `DeleteByQuery` is indeed faster.

Regarding your concern about the fact that `DeleteByQuery` is synchronous, this is expected as stated in the documentation. The write operations marked as asynchronous in the documentation are all the direct write calls to Algolia. Here, the `DeleteByQuery` is not a single atomic call to the API (it's a browse+delete objects), and to stay consistant with the documentation and other API clients, we have to keep it blocking.

Everything seems fine with your PR, I'll merge it. I'll just add a fix to the function as the tests seem not to pass on my side and tag the new version after that.

Thank you again for reporting the issue and helping us maintaining the Go client.

aseure:
Hello again @poweroftrue,

Just to keep you informed: your contribution is available as of version [2.16.0](https://github.com/algolia/algoliasearch-client-go/releases/tag/2.16.0) that I've deployed today.

Have a good day,

PR #336, opened 2017-08-23
Fix/maxidleconns

aseure:


algoliasearch/transport.go
```diff
@@ -434,3 +434,24 @@ func (t *Transport) setDialTimeout(dialTimeout time.Duration) {
...
+
+// setMaxIdleConns sets the `MaxIdleConns` (via the given `allHosts`) and
+// `MaxIdleConnsPerHost` (via the given `perHosts`) values of the underlying
+// RoundTripper of the HTTP client if it's an instance of `http.Transport`. If
+// one the given parameters is negative, it will get ignored and the underlying
+// values of the http.Transport will stay unchanged.
+func (t *Transport) setMaxIdleConns(allHosts, perHosts int) {
+	switch transport := t.httpClient.Transport.(type) {
+	case (*http.Transport):
+		if allHosts > -1 {
```
Nagriar: 0 is a big risky value because it's kind of the same than disabling the keepalive

aseure: Yes agree that it's dangerous. However, we are already letting the user completely override the transport layer. I'd feel the implementation would be a bit weird if we let the final user override only with certain values (non-zero as you're suggesting). WDYT?

Nagriar: fine with it

PR #353, opened 2017-10-10
Feat/synonyms and rules exporting

aseure:


algoliasearch/rules_iterator.go
```diff
@@ -0,0 +1,64 @@
...
+			return nil, err
+		}
+	}
+
+	it.pos++
+	if it.pos >= len(it.rules) {
+		return nil, NoMoreRulesErr
+	}
+
+	rule := it.rules[it.pos]
```
clement-leprovost: We should remove `_highlightResult` for rules too.

aseure:
@clement-leprovost You were right, I was missing the `_highlightResult` part of the answer. Implementing it right now and removing it from the `RuleIterator.Next()`.

PR #404, opened 2018-02-06
Add Indexes to Key

### LINKED ISSUE #403: Expose `indexes` in get API key / keys
Looks like this is present for API retrieve calls if set
python client:
```
c.get_api_key_acl('SECRET')
{u'description': u'Search-only API Key', u'validity': 0, u'indexes': [u'test-index-2'], u'value': u'SECRET', u'acl': [u'search'], u'createdAt': 1517349540}
```

bpicolo:
fixes #403, lmk if something is wrong here. Didn't see an existing test to add to


aseure:
Thank you for your contribution. This was indeed missing and stayed unnoticed for too long. Good catch 👍

PR #416, opened 2018-04-09
Add golang version to UserAgent

julienbourdeau:
`runtime.Version()` returns something like "go1.10.1", I believe it's best to remove the first 2 characters.

### DIFF
--- a/algoliasearch/transport.go
+++ b/algoliasearch/transport.go
@@ -12,6 +12,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"runtime"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


algoliasearch/transport.go
```diff
@@ -86,7 +87,7 @@ func NewTransportWithHosts(appId, apiKey string, hosts []string) *Transport {
 func defaultHeaders(appId, apiKey string) map[string]string {
 	return map[string]string{
 		"Connection":               "keep-alive",
-		"User-Agent":               "Algolia for Go (" + version + ")",
+		"User-Agent":               "Go ("+ runtime.Version()[2:] +"); Algolia for Go (" + version + ")",
```
aseure: 1. We don't want to panic if the string is shorter than 2 characters long. Why don't we keep the full runtime.Version() instead, which makes it complete, as @maxiloc suggested on Slack. It would let us use it with more flexibility on our side.

2. For clarity sake, let's use `fmt.Sprintf` to properly format the User-Agent here, such as:
```go
fmt.Sprintf("Go (%s); Algolia for Go (%s)", runtime.Version(), version)
```

julienbourdeau: I'm not sure we'll ever need the first 2 letter in `go1.10.1` but we definitely don't want to panic! I'm updated 👌

julienbourdeau:
I have checked in the dashboard and the client version must be passed first, I updated the code.

Please see https://gist.github.com/julienbourdeau/7c11df8647e85f5fc7eac9d2c017e52c

If the tests pass, the PR is ready to merge 🎉

julienbourdeau:
Tests on Travis: https://travis-ci.org/algolia/algoliasearch-client-go/builds/364184936
(something went wrong with GitHub)

PR #419, opened 2018-05-03
fix(key): Add QueryParameters in Key struct to fix typo

julienbourdeau:
There is a typo in QueryParameters. I am not sure this is the right way to fix it, there might be side effects that I haven't considered.

Feel free to update or close, or let me know what to do.

### DIFF
--- a/algoliasearch/types_key.go
+++ b/algoliasearch/types_key.go
@@ -7,7 +7,7 @@ type Key struct {
 	Indexes                []string `json:"indexes,omitempty"`
 	MaxHitsPerQuery        int      `json:"maxHitsPerQuery,omitempty"`
 	MaxQueriesPerIPPerHour int      `json:"maxQueriesPerIPPerHour,omitempty"`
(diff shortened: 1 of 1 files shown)


aseure:
This won't work as in Go, if you try to deserialize twice from the same JSON fields into two different structure fields, none of them will hold the value (eg. https://play.golang.org/p/LKRkujUG1LZ).

The only way would be to change the parameter name directly (in a breaking change release).

What I would suggest in to transform this PR into a `// TODO: XXX` comment so that we can postpone it for the release of the next major. WDYT @julienbourdeau ?

julienbourdeau:
Damn I didn't think Go will prevent you from using the same attribute twice. I'll update with a todo 👍

aseure:
As you and @nunomaduro recommended me, I've started a `develop` branch for this repository to put breaking changes (like this PR) without impacting the `master` branch that will still contain next non-breaking changes to be released on the current major version. This is why I've rebased this PR on `develop` and will merge it into it.

PR #420, opened 2018-05-03
Introduce Multi Cluster Management (ie MCM)

julienbourdeau:
Introduce Multi Cluster Management 🎉 

Any feedback is very welcome since I'm fairly new to Go.

- [x] Implement the 7 functions
- [x] Write the tests
- [x] Update documentation snippets (https://github.com/algolia/doc/pull/1537)

### DIFF
--- a/algoliasearch/algoliasearch.go
+++ b/algoliasearch/algoliasearch.go
@@ -197,6 +197,55 @@ type Client interface {
 	// also accepts extra RequestOptions.
 	MultipleQueriesWithRequestOptions(queries []IndexedQuery, strategy string, opts *RequestOptions) (res []MultipleQueryRes, err error)
 
+	// List all available clusters for the application.
+	ListClusters() (res []Cluster, err error)
+
+	// ListClustersWithRequestOptions is the same as ListClusters but it
+	// also accepts extra RequestOptions.
+	ListClustersWithRequestOptions(opts *RequestOptions) (res []Cluster, err error)
+
+	// List all user IDs across all clusters.
+	ListUserIDs(page int, hitsPerPage int) (res ListUserIDsRes, err error)
+
+	// ListUserIDsWithRequestOptions is the same as ListUserIDs but it
+	// also accepts extra RequestOptions.
+	ListUserIDsWithRequestOptions(page int, hitsPerPage int, opts *RequestOptions) (res ListUserIDsRes, err error)
+
+	// Get a specific user ID when using multi cluster infrastructure.
+	GetUserID(userID string) (res UserID, err error)
+
+	// GetUserIDWithRequestOptions is the same as GetUserID but it
+	// also accepts extra RequestOptions.
+	GetUserIDWithRequestOptions(userID string, opts *RequestOptions) (res UserID, err error)
+
+	// Assign a user ID to a cluster.
+	AssignUserID(userID string, clusterName string) (res AssignUserIDRes, err error)
+
+	// AssignUserIDWithRequestOptions is the same as AssignUserID but it
+	// also accepts extra RequestOptions.
+	AssignUserIDWithRequestOptions(userID string, clusterName string, opts *RequestOptions) (res AssignUserIDRes, err error)
+
+	// Remove a user ID from a cluster.
+	RemoveUserID(userID string) (res RemoveUserIDRes, err error)
+
+	// RemoveUserIDWithRequestOptions is the same as RemoveUserID but it
+	// also accepts extra RequestOptions.
+	RemoveUserIDWithRequestOptions(userID string, opts *RequestOptions) (res RemoveUserIDRes, err error)
+
+	// Get user IDs with the highest number of records per cluster.
+	GetTopUserIDs() (res TopUserIDs, err error)
+
+	// GetTopUserIDsWithRequestOptions is the same as GetTopUserIDs but it
+	// also accepts extra RequestOptions.
+	GetTopUserIDsWithRequestOptions(opts *RequestOptions) (res TopUserIDs, err error)
+
+	// Typical Algolia search but for user ID.
+	SearchUserIDs(query string, params Map) (res SearchUserIDRes, err error)
+
+	// SearchUserIDsWithRequestOptions is the same as SearchUserIDs but it
+	// also accepts extra RequestOptions.
+	SearchUserIDsWithRequestOptions(query string, params Map, opts *RequestOptions) (res SearchUserIDRes, err error)
+
 	// Batch performs all queries in `operations`.
 	Batch(operations []BatchOperationIndexed) (res MultipleBatchRes, err error)
 
--- a/algoliasearch/client.go
+++ b/algoliasearch/client.go
@@ -2,6 +2,7 @@ package algoliasearch
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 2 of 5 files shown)


algoliasearch/algoliasearch.go
```diff
@@ -172,6 +172,55 @@ type Client interface {
 	// also accepts extra RequestOptions.
 	MultipleQueriesWithRequestOptions(queries []IndexedQuery, strategy string, opts *RequestOptions) (res []MultipleQueryRes, err error)
 
+	// List all the cluster available for the application
```
aseure: `s/all the cluster available/all available clusters` and missing dot at the end.

algoliasearch/algoliasearch.go
```diff
@@ -172,6 +172,55 @@ type Client interface {
...
+	ListClusters() (res []Cluster, err error)
+
+	// ListClustersWithRequestOptions is the same as ListClusters but it
+	// also accepts extra RequestOptions.
+	ListClustersWithRequestOptions(opts *RequestOptions) (res []Cluster, err error)
+
+	// List all user IDs across all clusters
```
aseure: Missing dot at the end.

algoliasearch/algoliasearch.go
```diff
@@ -172,6 +172,55 @@ type Client interface {
...
+	ListUserIDs(page int, hitsPerPage int) (res listUserIDsRes, err error)
+
+	// ListUserIDsWithRequestOptions is the same as ListUserIDs but it
+	// also accepts extra RequestOptions.
+	ListUserIDsWithRequestOptions(page int, hitsPerPage int, opts *RequestOptions) (res listUserIDsRes, err error)
+
+	// Get a specific user ID
```
aseure: Missing dot at the end.

I also find the comment a bit poor in term of meaning (compared to the name of the method ;p), perhaps we should say somewhere it's related to MCM for instance.

algoliasearch/algoliasearch.go
```diff
@@ -172,6 +172,55 @@ type Client interface {
...
+	GetUserID(userID string) (res UserID, err error)
+
+	// GetUserIDWithRequestOptions is the same as GetUserID but it
+	// also accepts extra RequestOptions.
+	GetUserIDWithRequestOptions(userID string, opts *RequestOptions) (res UserID, err error)
+
+	// Assign a user ID to a cluster
```
aseure: Missing dot.

algoliasearch/algoliasearch.go
```diff
@@ -172,6 +172,55 @@ type Client interface {
...
+	AssignUserID(userID string, clusterName string) (res AssignUserIDRes, err error)
+
+	// AssignUserIDWithRequestOptions is the same as AssignUserID but it
+	// also accepts extra RequestOptions.
+	AssignUserIDWithRequestOptions(userID string, clusterName string, opts *RequestOptions) (res AssignUserIDRes, err error)
+
+	// Remove a user ID from a cluster
```
aseure: Missing `dot` and same as for `GetUserID`, we should perhaps add something in the comment saying this method is related to MCM.

algoliasearch/client.go
```diff
@@ -243,6 +243,92 @@ func (c *client) MultipleQueriesWithRequestOptions(queries []IndexedQuery, strat
...
 }
 
+func (c *client) ListClusters() (res []Cluster, err error) {
+	return c.ListClustersWithRequestOptions(nil)
+}
+
+func (c *client) ListClustersWithRequestOptions(opts *RequestOptions) (res []Cluster, err error) {
+	var tmp map[string][]Cluster
+	err = c.request(&tmp, "GET", "/1/clusters", nil, read, opts)
+	res = tmp["clusters"]
```
aseure: `tmp` is a bit smelly in term of code readiness. Use `res` instead as it's really the JSON response from the engine and use `clusters` as a new name for the returned `[]Cluster` variable.

algoliasearch/client.go
```diff
@@ -243,6 +243,92 @@ func (c *client) MultipleQueriesWithRequestOptions(queries []IndexedQuery, strat
...
+	err = c.request(&res, "GET", "/1/clusters/mapping?"+encodeMap(params), nil, read, opts)
+	return
+}
+
+func (c *client) GetUserID(userID string) (res UserID, err error) {
+	return c.GetUserIDWithRequestOptions(userID, nil)
+}
+
+func (c *client) GetUserIDWithRequestOptions(userID string, opts *RequestOptions) (res UserID, err error) {
+	err = c.request(&res, "GET", "/1/clusters/mapping/"+url.QueryEscape(userID), nil, read, opts)
```
aseure: Instead of directly passing the `userID` here, let's early return an error if it is empty. This way, we avoid an extra call to the engine and we can add a simple test for it.

algoliasearch/client.go
```diff
@@ -243,6 +243,92 @@ func (c *client) MultipleQueriesWithRequestOptions(queries []IndexedQuery, strat
...
+	return c.AssignUserIDWithRequestOptions(userID, clusterName, nil)
+}
+
+func (c *client) AssignUserIDWithRequestOptions(userID string, clusterName string, opts *RequestOptions) (res AssignUserIDRes, err error) {
+	if opts == nil {
+		opts = &RequestOptions{
+			ExtraHeaders: make(map[string]string),
+		}
+	}
+	opts.ExtraHeaders["X-Algolia-User-ID"] = userID
```
aseure: Following the previous comment, do we want to check for `userID` emptiness before sending the actual call? WDYT? If not, let's make sure it's not empty by early returning the error + adding a test.

algoliasearch/client_test.go
```diff
@@ -430,3 +432,75 @@ func TestDnsTimeout(t *testing.T) {
...
+	client := initMCMClient(t)
+	var uniqueUserID = fmt.Sprintf("go-MCM-tests-%d", rand.Int())
+
+	t.Log("TestMultiClusterManagement: Listing clusters")
+	clusters, err := client.ListClusters()
+	if err != nil {
+		t.Fatalf("TestMultiClusterManagement: Cannot list Clusters: %s", err)
+	}
+	if len(clusters) <= 1 {
+		t.Log("The app as 1 or less clusters, which is odd for an MCM app")
```
aseure: `s/as/has`

algoliasearch/client_test.go
```diff
@@ -430,3 +432,75 @@ func TestDnsTimeout(t *testing.T) {
...
+
+	t.Logf("TestMultiClusterManagement: Getting userID [%s]", uniqueUserID)
+	var userID UserID
+	for i := 0; i < 10; i++ {
+		userID, err = client.GetUserID(uniqueUserID)
+		if err == nil {
+			break
+		}
+		time.Sleep(time.Duration(i) * time.Second)
+	}
```
aseure: I'm not sure what we are testing here? Why are we doing 10 times the same test + adding sleep times in between. Not only this makes our tests take more time to run, but it's also weird that no error is checked in the end.

algoliasearch/client_test.go
```diff
@@ -430,3 +432,75 @@ func TestDnsTimeout(t *testing.T) {
...
+		t.Fatalf("TestMultiClusterManagement: Cannot get assigned userIDs: %s", err)
+	}
+
+	t.Log("TestMultiClusterManagement: Listing userIDs")
+	users, err := client.ListUserIDs(0, 10)
+	if err != nil {
+		t.Fatalf("TestMultiClusterManagement: Cannot list userIDs: %s", err)
+	}
+	if len(users.UserIDs) <= 1 {
+		t.Fatalf("The app has 1 or less clusters, which is odd considering we just added one")
```
aseure: Could you prefix the error with `TestMultiClusterManagement: ` (it's really easier for debugging later on, trust me ^^). Same goes for other places where it's missing, if any.

algoliasearch/client_test.go
```diff
@@ -430,3 +432,75 @@ func TestDnsTimeout(t *testing.T) {
 		t.Fatalf("TestDnsTimeout: Spent %d seconds instead of <5s to perform the 10 retries", int(delta.Seconds()))
 	}
 }
+
+func TestMultiClusterManagement(t *testing.T) {
+	client := initMCMClient(t)
+	var uniqueUserID = fmt.Sprintf("go-MCM-tests-%d", rand.Int())
```
aseure: Why do we need to introduce randomness here?

aseure: Note that I'm not super familiar with MCM and its userIDs.

algoliasearch/types_cluster.go
```diff
@@ -0,0 +1,8 @@
+package algoliasearch
+
+type Cluster struct {
+	ClusterName string `json:"clusterName"`
+	NbRecords   int    `json:"nbRecords"`
+	NbUserIDs   int    `json:"nbUserIDs"`
+	DataSize    int    `json:"dataSize"`
+}
```
aseure: Why do you think of putting all the types related to MCM under `types_mcm.go`? I find it clearer IMHO.

julienbourdeau: Definitely 👍

algoliasearch/client_test.go
julienbourdeau: Unfortunately, MCM doesn't return taskID, so if the test is too fast, the userID is not yet available when we try to get it. I added this loop to try mulitple times and wait. In the end, if no userID could be found after the 10th try, it will file line 467.

algoliasearch/client_test.go
julienbourdeau: 😅

algoliasearch/client_test.go
julienbourdeau: While developing, tests would often fail and relaunching would also fail because of passed userID already existing ^^
It's not so important once everything is stable but it really helps you have "operation pending on userId XX errors".

algoliasearch/client.go
julienbourdeau: Yeah, agreed 👍

algoliasearch/client_test.go
aseure: I don't know if it's possible with MCM but what I'm doing for all the other tests is that I force clean/remove all the possible resources I'm gonna use (usually I delete the index and recreate it right after for instance) so that I'm 100% sure my test is always running the same, whether or not there were leftovers from previous tests. WDYT?

algoliasearch/client_test.go
aseure: I don't think we want our users to do the same as you did here with a wait etc. What about making the `GetUserID` synchronous then, and encapsulate the retry internally, with a fix number of retries and finally return the error?

PR #424, opened 2018-06-15
feat(client): Expose waitTask in the client

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | no     
| Need Doc update   | yes (not sure)

## What was changed

The 2 methods about taskID initially available in the `Index` moved to the `Client`. The original methods on the index still work are **not** deprecated.

## Why it was changed

You could get some taskID from the engine without necessarily having an instance of Index. Instead of instantiating an index that you won't need, you can now call `QaitTask` and `GetStatus` on the client.

* It will be very useful for the new Analytics object (see https://github.com/algolia/algoliasearch-client-php/pull/408)
* We're planning to merge https://github.com/algolia/algoliasearch-client-php/pull/295 (and port it to all clients). Adding multipleObjects will be easier to use if QaitTask lives in the Client.
* It was also discussed before for the Django integration ([here](https://github.com/algolia/algoliasearch-django/pull/258)).

### DIFF
--- a/algoliasearch/client.go
+++ b/algoliasearch/client.go
@@ -2,6 +2,7 @@ package algoliasearch
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/url"
 	"time"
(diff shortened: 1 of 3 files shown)


algoliasearch/client.go
```diff
@@ -258,6 +259,48 @@ func (c *client) BatchWithRequestOptions(operations []BatchOperationIndexed, opt
...
+func (c *client) WaitTask(indexName string, taskID int) error {
+	return c.WaitTaskWithRequestOptions(indexName, taskID, nil)
+}
+
+func (c *client) WaitTaskWithRequestOptions(indexName string, taskID int, opts *RequestOptions) error {
+	var res TaskStatusRes
+	var err error
+
+	var maxDuration = time.Second
+	var sleepDuration time.Duration
```
aseure: As you do not reuse the variables outside of the `for` loop, I'd recommend you to remove all those lines, replace the `res, err =  c.GetStatusWithRequestOptions(...)` with  `res, err := c.GetStatusWithRequestOptions(...)` and directly do `sleepDuration := randDuration(time.Second)` in your `for` loop. It would save 5 useless lines.

algoliasearch/client.go
```diff
@@ -258,6 +259,48 @@ func (c *client) BatchWithRequestOptions(operations []BatchOperationIndexed, opt
...
+			return nil
+		}
+
+		sleepDuration = randDuration(maxDuration)
+		time.Sleep(sleepDuration)
+
+		// Increase the upper boundary used to generate the sleep
+		// duration
+		if maxDuration < 10*time.Minute {
+			maxDuration *= 2
```
aseure: As we probably want to be bounded by 10min, why do you think of capping the value of `maxDuration` to 10min just after this line? Otherwise, if `maxDuration` was at 6min for instance, you would have to wait for 12min afterwards.

algoliasearch/client.go
julienbourdeau: To be honest, I copypasted what was in `index.go`, I didn't even read the function :D

PR #428, opened 2018-06-20
chore: Add release.sh script

aseure:
Adding a `release.sh` script to perform a release in a single step.

What it does is to:
1. Update the version of the client by changing it in the `algoliasearch/transport.go` file
2. Update the `ChangeLog.md` using our ChangeLog generator tool
3. Ask the user for confirmation of the changes to be commited
4. Commit, tag and push on the master branch

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+set -e
+
+function is_version_valid() {
+  echo "$1" | grep -q -E '^[0-9]+\.[0-9]+\.[0-9]+$'
+  if [[ "$?" == "0" ]]; then return 0; else return 1; fi
+}
+
+function is_working_dir_clean() {
+  if [ -z "$(git status --porcelain)" ]; then return 0; else return 1; fi
+}
+
+version=$1
+
+if ! is_version_valid "$version"; then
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+set -e
+
+function is_version_valid() {
+  echo "$1" | grep -q -E '^[0-9]+\.[0-9]+\.[0-9]+$'
+  if [[ "$?" == "0" ]]; then return 0; else return 1; fi
+}
+
+git reset --hard HEAD > /dev/null 2>&1
```
rayrutjes: This appears a bit harsh.
It happens to me sometimes that I made some changes that I do not want to lose, and figure it out the moment I release.

Maybe aborting the releasing, and pin pointing the fact that the current state is not clean is better?

julienbourdeau: We could also stash at the beginning and stash pop at the end.

aseure: I'd rather not stash in scripts, I had bad surprises in the past.

I'd go for either:
1. Raymond' solution where we just abort if the worktree isn't clear
2. Just reset the `transport.go` and `ChangeLog.md` files as those are the only one we modify.

WDYT @rayrutjes @julienbourdeau? 1. looks safer to me indeed.

PR #432, opened 2018-07-25
Query rules v2

aseure:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | 
| Need Doc update   | yes

## What was changed

All changes related to the release of Query Rules v2. For detailled informations, please take a look at the dedicated commit. Other commits simply add the associated tests and introduce a new dependency on github.com/mitchellh/mapstructure now used in tests only to ease response objects comparison.

## Why it was changed

New features available from the engine.

### DIFF
--- a/algoliasearch/types_rule.go
+++ b/algoliasearch/types_rule.go
@@ -1,11 +1,35 @@
 package algoliasearch
 
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+)
+
 type Rule struct {
-	ObjectID        string          `json:"objectID,omitempty"`
-	Condition       RuleCondition   `json:"condition"`
-	Consequence     RuleConsequence `json:"consequence"`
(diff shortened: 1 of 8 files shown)


algoliasearch/types_rule.go
```diff
@@ -1,11 +1,35 @@
...
+	ObjectID             string          `json:"objectID,omitempty"`
+	Validity             []TimeRange     `json:"validity,omitempty"`
+	isExplicitlyDisabled bool
+}
+
+func (r *Rule) Enable() {
+	r.Enabled = true
+}
+
+func (r *Rule) Disable() {
```
julienbourdeau: After discussing it, I guess having these `Enable` / `Disable` methods is the best way to handle it. We'll need to document it tho.

PR #433, opened 2018-08-06
fix: objectID is required to deleteObject

julienbourdeau:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Need Doc update   | no

## What was changed

`Index::delete_object` needs a non-empty objectID

## Why it was changed

If the object_id is empty, the entire index will be deleted instead of one record.


aseure:
As I'm clearing the long-running PR of the Go client, I took the opportunity to finish this one @julienbourdeau ;)

PR #435, opened 2018-08-22
fix: Improve retry strategy by reseting default hosts after some time

aseure:
We recently discovered that the implementation of the retry strategy for
the client was lacking an important feature: the ability to reset hosts
marked as down after some time. Because of that, an host that was marked
down could never be retried without instantiating a new client.

This commit is a complete rewrite of the retry strategy implementation
to address this issue. For simplicity, we decoupled the retry strategy
implementation from the transport layer, responsible for issuing the
HTTP calls. This new implementation was written in a concurrent-safe
way, meaning that concurrent calls should not result in UB regarding the
state of the retry strategy's hosts.


algoliasearch/retry_strategy.go
```diff
@@ -0,0 +1,266 @@
...
+	for _, h := range s.hosts {
+		if h.accept(k) {
+			h.reset()
+			hosts = append(hosts, &tryableHost{h.host, baseTimeout})
+		}
+	}
+	return hosts
+}
+
+func (s *retryStrategy) Decide(h TryableHost, code int, err error) Outcome {
```
julienbourdeau: 👍

algoliasearch/transport.go
```diff
@@ -30,293 +31,88 @@ const (
...
-			t.resetDialTimeout()
-			if typeCall == write {
-				t.activeWriteSince = time.Now()
-				t.activeWriteHost = host
-			} else {
-				t.activeReadSince = time.Now()
-				t.activeReadHost = host
-			}
-			return res, nil
+	for _, h := range t.retryStrategy.GetTryableHosts(k) {
```
julienbourdeau: 👍

aseure:
@julienbourdeau Yes only breaking change is the drop of Go 1.6 because we are now using contexts. The badge in the `README.md` and the Travis build have been changed.

PR #446, opened 2018-10-29
changed: Ensure that after a index.move, the internal name of the index is correctly updated as well

aseure:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | N/A
| Need Doc update   | no

### DIFF
--- a/algoliasearch/index.go
+++ b/algoliasearch/index.go
@@ -406,7 +406,11 @@ func (i *index) Move(name string) (UpdateTaskRes, error) {
 }
 
 func (i *index) MoveWithRequestOptions(name string, opts *RequestOptions) (UpdateTaskRes, error) {
-	return i.operation(name, "move", nil, opts)
+	res, err := i.operation(name, "move", nil, opts)
+	if err == nil {
+		i.name = name
+	}
+	return res, err
 }
 
 func (i *index) operation(dst, op string, scopes []string, opts *RequestOptions) (res UpdateTaskRes, err error) {


algoliasearch/index.go
```diff
@@ -406,7 +406,11 @@ func (i *index) Move(name string) (UpdateTaskRes, error) {
 }
 
 func (i *index) MoveWithRequestOptions(name string, opts *RequestOptions) (UpdateTaskRes, error) {
-	return i.operation(name, "move", nil, opts)
+	res, err := i.operation(name, "move", nil, opts)
```
nunomaduro: Unfortunately this may be a breaking change. The index object no longer targets the old index, but the new one.

Shouldn't we consider move this to the v2 version?

nunomaduro: @aseure Unless we consider this a bug 😁

aseure: Arghhh, you're right. Good catch. I'll base this PR on the `develop` branch then (next major release branch). Thank you.

aseure: @nunomaduro At first, I wanted to consider it as a bug but you're right, it's a breaking change as one may have relied on this behaviour with this scenario for instance:

1. index := client.InitIndex("old_name")
2. index.AddObjects(/* pass 10 objects */)
3. index.Move("new_name")
4. index.AddObjects(/* pass 20 objects */)

In that case, "old_name" index contains the latest 20 objects and "new_name" contains the first 10 objects. However, with the proposed change, we would have all 30 objects on "new_name" index.

I've just made `develop` branch the new base branch of this PR so that it will only be included with the upcoming v3 of the client to avoid breaking the compatibility. Thanks for your thorough review.

nunomaduro: @aseure Nice, this is definitively this safest move. 👍

PR #452, opened 2018-11-08
Offsite changes

aseure:
* added: Implement AccountClient.CopyIndex 
 * changed: remove extra empty line 
 * changed: Use the environment variables from our Common Test Suite for testing 
 * added: Implement Index.GetAppID and Client.GetAppID methods 
 * added: Implement Index.Replace{Objects,Rules,Synonyms} 
 * deprecated: Index.{Delete,Clear} replaced by Client.{Delete,Clear}Index 
 * added: Implement Client.Copy{Settings,Synonyms,Rules} 
 * deprecated: Index.{Copy,Move} replaced by Client.{Copy,Move}Index

### DIFF
--- a/algoliasearch/account_client.go
+++ b/algoliasearch/account_client.go
@@ -1,0 +1,135 @@
+package algoliasearch
+
+import "fmt"
+
+type accountClient struct{}
+
+func NewAccountClient() AccountClient {
+	return &accountClient{}
+}
+
+func (a *accountClient) CopyIndex(src, dst Index) ([]int, error) {
+	return a.CopyIndexWithRequestOptions(src, dst, nil)
+}
+
+func (a *accountClient) CopyIndexWithRequestOptions(src, dst Index, opts *RequestOptions) ([]int, error) {
+	if src.GetAppID() == dst.GetAppID() {
+		return nil, SameAppIDErr
+	}
+
+	if _, err := dst.GetSettingsWithRequestOptions(opts); err == nil {
+		return nil, IndexAlreadyExistsErr
+	}
+
+	var taskIDs []int
+
+	// Copy synonyms
+	{
+		it := NewSynonymIterator(src)
+
+		var synonyms []Synonym
+
+		for {
+			synonym, err := it.Next()
+			if err != nil {
+				if err == NoMoreSynonymsErr {
+					break
+				} else {
+					return nil, fmt.Errorf("error while iterating source index synonyms: %v", err)
+				}
+			}
+			synonyms = append(synonyms, *synonym)
+		}
+
+		if synonyms != nil {
+			res, err := dst.ReplaceAllSynonymsWithRequestOptions(synonyms, opts)
+			if err != nil {
+				return nil, fmt.Errorf("error while replacing destination index synonyms: %v", err)
+			}
+			taskIDs = append(taskIDs, res.TaskID)
+		}
+	}
+
+	// Copy rules
+	{
+		it := NewRuleIterator(src)
+
+		var rules []Rule
+
+		for {
+			rule, err := it.Next()
+			if err != nil {
+				if err == NoMoreRulesErr {
+					break
+				} else {
+					return nil, fmt.Errorf("error while iterating source index rules: %v", err)
+				}
+			}
+			rules = append(rules, *rule)
+		}
+		if rules != nil {
+			res, err := dst.ReplaceAllRulesWithRequestOptions(rules, opts)
+			if err != nil {
+				return nil, fmt.Errorf("error while replacing destination index rules: %v", err)
+			}
+			taskIDs = append(taskIDs, res.TaskID)
+		}
+	}
+
+	// Copy settings
+	{
+		settings, err := src.GetSettingsWithRequestOptions(opts)
+		if err != nil {
+			return nil, fmt.Errorf("cannot retrieve source index settings: %v", err)
+		}
+
+		res, err := dst.SetSettingsWithRequestOptions(settings.ToMap(), opts)
+		if err != nil {
+			return nil, fmt.Errorf("cannot set destination index settings: %v", err)
+		}
+		taskIDs = append(taskIDs, res.TaskID)
+
+	}
+
+	// Copy objects
+	{
+		it, err := src.BrowseAllWithRequestOptions(nil, opts)
+		if err != nil {
+			return nil, fmt.Errorf("cannot browse source index objects: %v", err)
+		}
+
+		var objects []Object
+		batchSize := 1000
+
+		for {
+			res, err := it.Next()
+			if err != nil {
+				if err == NoMoreHitsErr {
+					break
+				} else {
+					return nil, fmt.Errorf("error while iterating source index objects: %v", err)
+				}
+			}
+			objects = append(objects, Object(res))
+
+			if len(objects) >= batchSize {
+				res, err := dst.AddObjectsWithRequestOptions(objects, opts)
+				if err != nil {
+					return nil, fmt.Errorf("error while saving batch of objects: %v", err)
+				}
+				taskIDs = append(taskIDs, res.TaskID)
+				objects = []Object{}
+			}
+		}
+
+		// Send the last batch
+		res, err := dst.AddObjectsWithRequestOptions(objects, opts)
+		if err != nil {
+			return nil, fmt.Errorf("error while saving batch of objects: %v", err)
+		}
+		taskIDs = append(taskIDs, res.TaskID)
+		objects = []Object{}
+	}
+
+	return taskIDs, nil
+}
--- a/algoliasearch/algoliasearch.go
+++ b/algoliasearch/algoliasearch.go
@@ -816,6 +816,27 @@ type Index interface {
 	// SearchRulesWithRequestOptions is the same as SearchRules but it also
 	// accepts extra RequestOptions.
 	SearchRulesWithRequestOptions(params Map, opts *RequestOptions) (SearchRulesRes, error)
+
+	// ReplaceAllSynonyms replace all the synonyms of the current index with the given ones.
+	ReplaceAllSynonyms(synonyms []Synonym) (res UpdateTaskRes, err error)
+
+	// ReplaceAllSynonymsWithRequestOptions is the same as ReplaceAllSynonyms but it also
+	// accepts extra RequestOptions.
+	ReplaceAllSynonymsWithRequestOptions(synonyms []Synonym, opts *RequestOptions) (res UpdateTaskRes, err error)
+
+	// ReplaceAllRules replace all the rules of the current index with the given ones.
+	ReplaceAllRules(rules []Rule) (res BatchRulesRes, err error)
+
+	// ReplaceAllRulesWithRequestOptions is the same as ReplaceAllRules but it also
+	// accepts extra RequestOptions.
+	ReplaceAllRulesWithRequestOptions(rules []Rule, opts *RequestOptions) (res BatchRulesRes, err error)
+
+	// ReplaceAllObjects replace all the objects of the current index with the given ones.
+	ReplaceAllObjects(objects []Object) error
+
+	// ReplaceAllObjectsWithRequestOptions is the same as ReplaceAllObjects but it also
+	// accepts extra RequestOptions.
+	ReplaceAllObjectsWithRequestOptions(objects []Object, opts *RequestOptions) error
 }
 
 // IndexIterator is used by the BrowseAll functions to iterate over all the
--- a/algoliasearch/algoliasearch.go
+++ b/algoliasearch/algoliasearch.go
@@ -9,6 +9,9 @@ import (
 // allows manipulations over the indexes of the application as well as network
 // related parameters.
 type Client interface {
+	// GetAppID returns the Algolia application ID to which the client is linked to.
+	GetAppID() string
+
 	// SetExtraHeader allows to set custom headers while reaching out to
 	// Algolia servers.
 	SetExtraHeader(key, value string)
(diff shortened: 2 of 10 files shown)


algoliasearch/account_client.go
```diff
@@ -0,0 +1,123 @@
+package algoliasearch
+
+import "fmt"
+
+type accountClient struct{}
+
+func NewAccountClient() AccountClient {
+	return &accountClient{}
+
```
nunomaduro: extra line here.

algoliasearch/account_client.go
```diff
@@ -0,0 +1,123 @@
...
+		for {
+			res, err := it.Next()
+			if err != nil {
+				if err == NoMoreHitsErr {
+					break
+				} else {
+					return fmt.Errorf("error while iterating source index records: %v", err)
+				}
+			}
+			records = append(records, Object(res))
```
nunomaduro: Unless I am missing something here, there is an algorithm problem in your code. You are not leveraging the fact that `it` is an iterator -  and your records variable here may cause a `memory exhausted`. You should try todo something similar to:

```python
        # Copy objects
        responses = []
        batch = []
        batch_size = 1000
        count = 0
        for obj in source_index.browse_all():
            batch.append(obj)
            count += 1

            if count == batch_size:
                response = destination_index.save_objects(batch, request_options)
                responses.append(response)
                batch = []
                count = 0

        if batch:
            response = destination_index.save_objects(batch, request_options)
            responses.append(response)

        return responses
```

algoliasearch/account_client.go
```diff
@@ -0,0 +1,123 @@
...
+import "fmt"
+
+type accountClient struct{}
+
+func NewAccountClient() AccountClient {
+	return &accountClient{}
+
+}
+
+func (a *accountClient) CopyIndex(src, dst Index) (err error) {
```
nunomaduro: Shouldn't we add `request_options` here?

algoliasearch/account_client.go
```diff
@@ -0,0 +1,123 @@
...
+
+	{
+		res, err := dst.SetSettings(settings.ToMap())
+		if err != nil {
+			return fmt.Errorf("cannot set destination index settings: %v", err)
+		}
+		taskIDs = append(taskIDs, res.TaskID)
+	}
+
+	for _, taskID := range taskIDs {
```
nunomaduro: The developer here don't have the possibility to choosing if he wants to wait or not. Can't you return `taskIDs` variable and the developer can wait afterwords if he wants?

algoliasearch/account_client.go
```diff
@@ -0,0 +1,123 @@
...
+func (a *accountClient) CopyIndex(src, dst Index) (err error) {
+	if src.GetAppID() == dst.GetAppID() {
+		return SameAppIDErr
+	}
+
+	if _, err := dst.GetSettings(); err == nil {
+		return IndexAlreadyExistsErr
+	}
+
+	// Copy objects
```
nunomaduro: IMO: The objects should be the last thing to add. Why? Because if you add the objects now, when you `rules` a big re-indexation will happen at Algolia side, when you add `synonyms` a big re-indexation will happen at Algolia side and also when you add `settings` a big re-indexation will happen at Algolia side.

A good solution may be just leave the add objects to the end of the method.

algoliasearch/algoliasearch.go
```diff
@@ -9,6 +9,9 @@ import (
 // allows manipulations over the indexes of the application as well as network
 // related parameters.
 type Client interface {
+	// GetAppID returns the Algolia application ID to which the client is linked to.
+	GetAppID() string
```
nunomaduro: This is breaking, we already had this discussing before. Just wanted it again, in order to consider move this Pull Request to the go v3.

algoliasearch/errors.go
```diff
@@ -10,6 +10,8 @@ var (
 	NoMoreSynonymsErr           error = errors.New("No more synonyms")
 	NoMoreRulesErr              error = errors.New("No more rules")
 	ExhaustionOfTryableHostsErr error = errors.New("All hosts have been contacted unsuccessfully")
+	SameAppIDErr                error = errors.New("Application IDs cannot be identical")
```
nunomaduro: IMO a better description: "The given indexes are on the same application. Use client.copy_index instead."

algoliasearch/errors.go
```diff
@@ -10,6 +10,8 @@ var (
...
+	IndexAlreadyExistsErr       error = errors.New("Index already exists")
```
nunomaduro: "Destination index already exists. Please delete it before copying index across applications."

algoliasearch/index.go
```diff
@@ -770,3 +759,87 @@ func (i *index) SearchRulesWithRequestOptions(params Map, opts *RequestOptions)
...
+	}()
+
+	var taskIDs []int
+
+	tmpIndex := i.client.InitIndex(tmpIndexName)
+	batchSize := 1000
+
+	for i := 0; i < len(objects); i += batchSize {
+		j := i + batchSize
+		if j > len(objects) {
```
nunomaduro: Performance issue ( at least on PHP ): You may want to increase a counter instead of calculating the size of the objects in every iteration. Example:
```python

        batch = []
        batch_size = 1000
        count = 0
        for obj in objects:
            batch.append(obj)
            count += 1

            if count == batch_size:
                response = tmp_index.save_objects(batch, request_options)
                responses.append(response)
                batch = []
                count = 0

        if batch:
            response = tmp_index.save_objects(batch, request_options)
            responses.append(response)
```

algoliasearch/index.go
```diff
@@ -770,3 +759,87 @@ func (i *index) SearchRulesWithRequestOptions(params Map, opts *RequestOptions)
...
+}
+
+// ReplaceAllObjects replace all the objects of the current index with the given ones.
+func (i *index) ReplaceAllObjects(objects []Object) error {
+	return i.ReplaceAllObjectsWithRequestOptions(objects, nil)
+}
+
+// ReplaceAllObjectsWithRequestOptions is the same as ReplaceAllObjects but it also
+// accepts extra RequestOptions.
+func (i *index) ReplaceAllObjectsWithRequestOptions(objects []Object, opts *RequestOptions) (err error) {
```
nunomaduro: Should we add the possibility of having a `safe` parameter?

algoliasearch/account_client.go
aseure: This is an interesting point. I thought that I would collect all the records in memory, which indeed may lead to a high memory footprint, depending on the index' total number of records, to leverage the `ReplaceAllObjects` that performs the automatic batching (1k objects at a time). As it's an helper method, I wanted to keep it simple and not introduce too much complexity, hence my use of the `replaceAll*` methods with the total number of objects. Do you think it's still unreasonable? cc @Ant-hem what are you thoughts on this?

algoliasearch/account_client.go
aseure: You're right. As discussed, fixing it by returning the list of taskIDs to let the user decide whether or not to wait on them. A `waitable`-like object will rather come in the next major release to make the client easier to use.

algoliasearch/account_client.go
rayrutjes: IMHO all the features we provide should be scalable.

algoliasearch/algoliasearch.go
aseure: As discussed, using an interface was a bad idea. Without changing the interface, I can't really touch any part of the public API (add/delete/modify methods). So for the current major version of the client, decision was made to let us break the compatibility for added methods only.

However, I definitely hear your point on compatibility and we'll make sure to do differently in the next major release.

algoliasearch/errors.go
aseure: Indeed, thank you for your careful feedback. Super good point 👍

algoliasearch/account_client.go
Ant-hem: @aseure  I think it's good idea to split the records. I have done this in Java and C#V6 for three main reasons:  memory, network and scaling. It's also great because our users don't have to worry about their lists size.

IMHO, next gen client should implement this behavior by default for `save*` methods.

algoliasearch/index.go
aseure: (not sure if the code snippet you share do match the comment you wrote)

Retrieving the length of a *slice* in Go (vector-like in other runtimes) is O(1) so we're fine here.

algoliasearch/index.go
aseure: As safe is not handled at all so far and I'm planning on starting to integrate it in the coming next major, I would again delay this in the next major. Let me know if you think otherwise @Ant-hem @nunomaduro, there's no technical blocking point to implement it now.

algoliasearch/account_client.go
aseure: As discussed with all of you, we will rather *not* use the `ReplaceAllObjects` here (to avoid useless temporary index) and we'll send the batch every time it reaches the batch size to avoid consuming too much memory. Thank you all 👍

algoliasearch/index.go
aseure: As discussed, because the change is breaking the compatibility (with regards to `RequestOptions`) and because we prefer to implement a new design for the `RequestOptions` in the next major release, we will not handle the optional `safe` parameter in this PR.

algoliasearch/account_client.go
```diff
@@ -0,0 +1,135 @@
...
+					break
+				} else {
+					return nil, fmt.Errorf("error while iterating source index synonyms: %v", err)
+				}
+			}
+			synonyms = append(synonyms, *synonym)
+		}
+
+		if synonyms != nil {
+			res, err := dst.ReplaceAllSynonyms(synonyms)
```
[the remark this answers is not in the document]
aseure: Good question. I don't really know actually. I always thought about `RequestOptions` as parameters used in the top-most function but we may want to propagate them in the source and/or destination client. What is your take on that?

algoliasearch/index.go
```diff
@@ -770,3 +759,87 @@ func (i *index) SearchRulesWithRequestOptions(params Map, opts *RequestOptions)
...
+
+// ReplaceAllSynonyms replace all the synonyms of the current index with the given ones.
+func (i *index) ReplaceAllSynonyms(synonyms []Synonym) (res UpdateTaskRes, err error) {
+	return i.ReplaceAllSynonymsWithRequestOptions(synonyms, nil)
+}
+
+// ReplaceAllSynonymsWithRequestOptions is the same as ReplaceAllSynonyms but it also
+// accepts extra RequestOptions.
+func (i *index) ReplaceAllSynonymsWithRequestOptions(synonyms []Synonym, opts *RequestOptions) (res UpdateTaskRes, err error) {
+	return i.BatchSynonymsWithRequestOptions(synonyms, true, true, opts)
```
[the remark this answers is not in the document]
aseure: True. I'll set it to false and make it replaceable using `RequestOptions` in the v3.

algoliasearch/index.go
```diff
@@ -770,3 +759,87 @@ func (i *index) SearchRulesWithRequestOptions(params Map, opts *RequestOptions)
...
+
+// ReplaceAllRules replace all the rules of the current index with the given ones.
+func (i *index) ReplaceAllRules(rules []Rule) (res BatchRulesRes, err error) {
+	return i.ReplaceAllRulesWithRequestOptions(rules, nil)
+}
+
+// ReplaceAllRulesWithRequestOptions is the same as ReplaceAllRules but it also
+// accepts extra RequestOptions.
+func (i *index) ReplaceAllRulesWithRequestOptions(rules []Rule, opts *RequestOptions) (res BatchRulesRes, err error) {
+	return i.BatchRulesWithRequestOptions(rules, true, true, opts)
```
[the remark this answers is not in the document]
aseure: True. I'll set it to false and make it replaceable using `RequestOptions` in the v3.

algoliasearch/account_client.go
```diff
@@ -0,0 +1,135 @@
...
+			if err != nil {
+				if err == NoMoreHitsErr {
+					break
+				} else {
+					return nil, fmt.Errorf("error while iterating source index objects: %v", err)
+				}
+			}
+			objects = append(objects, Object(res))
+
+			if len(objects) > batchSize {
```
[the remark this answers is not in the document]
aseure: True. Let's avoid tickling anyone subject to OCD there 😂

algoliasearch/account_client.go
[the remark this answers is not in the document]
nunomaduro: @aseure That's a good point. IMO: The request option object may contain information useful for the http layer, such as: `headers`, etc. I think the developer should be able to have the hand on that. cc @Ant-hem @julienbourdeau.

Side note: On PHP and Python we are passing requestOptions to all http calls.

algoliasearch/account_client.go
```diff
@@ -0,0 +1,135 @@
...
+				if err != nil {
+					return nil, fmt.Errorf("error while saving batch of objects: %v", err)
+				}
+				taskIDs = append(taskIDs, res.TaskID)
+				objects = []Object{}
+			}
+		}
+
+		// Send the last batch
+		res, err := dst.AddObjects(objects)
```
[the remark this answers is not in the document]
aseure: I'd rather address this issue in the `saveObjects` directly when we will fix the issue. Good catch 👍

algoliasearch/index.go
```diff
@@ -770,3 +759,87 @@ func (i *index) SearchRulesWithRequestOptions(params Map, opts *RequestOptions)
...
+		res, err = tmpIndex.AddObjectsWithRequestOptions(objects[i:j], opts)
+		if err != nil {
+			return
+		}
+		taskIDs = append(taskIDs, res.TaskID)
+	}
+
+	{
+		var res UpdateTaskRes
+		res, err = i.client.ScopedCopyIndexWithRequestOptions(i.name, tmpIndexName, []string{"settings", "synonyms", "rules"}, opts)
```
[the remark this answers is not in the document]
aseure: Again, you're right. Thank you for catching this one.

PR #465, opened 2019-02-06
added: implement Client.RestoreAPIKey

aseure:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | 
| Need Doc update   | no

## Describe your change

Implement the new `Client.RestoreAPIKey` method.

### DIFF
--- a/algoliasearch/client_test.go
+++ b/algoliasearch/client_test.go
@@ -191,6 +191,33 @@ func TestClientKeys(t *testing.T) {
 
 		waitClientKey(t, c, searchKey, func(k Key) bool { return k.Description == description })
 	}
+
+	t.Log("TestClientKeys: Delete search key and restore it")
+	{
+		_, err := c.DeleteAPIKey(searchKey)
+		require.NoError(t, err)
+
+		// Wait until the key has been properly deleted
+		for {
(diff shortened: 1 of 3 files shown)


algoliasearch/client_test.go
```diff
@@ -191,6 +191,33 @@ func TestClientKeys(t *testing.T) {
...
+
+		// Wait until the key has been properly restored
+		for {
+			_, err := c.GetAPIKey(searchKey)
+			if err == nil {
+				break
+			}
+			time.Sleep(1*time.Second)
+		}
+	}
```
Ant-hem: It could be nice to delete the key afterward, to avoid having garbage in indices.

aseure: Temporary keys are already deleted at the end of the function' scope thanks to the two `defer` calls:
 * https://github.com/algolia/algoliasearch-client-go/blob/3688ec464a24c08a09a3663ac466353410196550/algoliasearch/client_test.go#L155
 * https://github.com/algolia/algoliasearch-client-go/blob/3688ec464a24c08a09a3663ac466353410196550/algoliasearch/client_test.go#L178

PR #468, opened 2019-03-14
Fix function comments based on best practices from Effective Go

CodeLingoTeam:
Hi, we updated an exported function comment based on best practices from [Effective Go](https://golang.org/doc/effective_go.html). It’s admittedly a relatively minor fix up. Does this help you?


aseure:
Thank you for contributing! We plan on documenting all the missing methods as well but this is still a good catch!

aseure:
Oups. I've just saw that your PR is only affecting the previous `v2` of our client. As we've just released a new `v3.0.0-beta01` version, this issue is not relevant to `master` anymore.

Let's open a new PR on a `v2` branch first.

PR #478, opened 2019-05-08
Fix/user data

ElPicador:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | yes    
| Need Doc update   | no

## Describe your change

`userData` in the query response is an array, see: https://www.algolia.com/doc/api-reference/api-methods/search/?language=javascript#method-response-userdata

see above


aseure:
@ElPicador I've rebased/pushed your branch to fix the conflict with the duplicated commit.

Thanks for all the fixes. In case you missed it, there's a way to deserialize the `UserData` field to your custom type. On the response object, instead of doing:

```go
var s []interface
s := QueryRes.UserData
// Do something with s
```

You can use the `UnmarshalUserData` method:

```go
var s YourCustomUserDataType
err := queryRes.UnmarshalUserData(&s)
// Do something with s
```

PR #482, opened 2019-06-07
Add check for ObjectIDProvider interface

pavel-popov:
| Q                 | A
| Bug fix?          | no
| New feature?      | no 
| BC breaks?        | no     
| Related Issue     | Fix #...  
| Need Doc update   | yes

## Describe your change

 Check if object implements `ObjectIDProvider` interface, if yes then respect it
and don't do heavy `json.Marshal`/`json.Unmarshal` operations.

In the `utils.getObjectID()` function we do `json.Marshal()` and `json.Unmarshal`. Then in `buildRequest` we do `json.Marshal` again. This can be pretty time consuming.

### DIFF
--- a/algolia/search/utils.go
+++ b/algolia/search/utils.go
@@ -53,7 +53,17 @@ func waitWithRetry(f func() (bool, error)) error {
 	}
 }
 
+// ObjectIDProvider can generate object ID.
+type ObjectIDProvider interface {
+	ObjectID() string
+}
+
 func getObjectID(object interface{}) (string, bool) {
+	o, ok := object.(ObjectIDProvider)
+	if ok {
+		return o.ObjectID(), true
+	}
+
 	data, err := json.Marshal(object)
 	if err != nil {
 		return "", false


aseure:
Hello @pavel-popov. Thank you for your contribution.

The `utils.getObjectID()` function can indeed be improved. Two things that could be done at the moment:
 1. expose a public interface (as you did) for users to implement
 2. use reflection to detect `objectID` JSON tags

We would like to move forward with one or both of these features in the future, hence your PR is really welcome.

However, we will need to leave it open for now and discuss/merge it later (as it won't be a breaking change anyway) because we plan to release the `v3.0.0` today and since the public API is frozen at the moment, we cannot expose a new public interface yet.

We'll come back to you soon to address this rightfully reported performance issue. In the meantime, if you can find the time, it would really help to add some benchmarks here to see the improvement.

Thank you very much for your contribution. Talk to you soon.

aseure:
Closing in favor of #500.

@pavel-popov don't hesitate to comment on the other PR which contains both your implementation and another one, as well as benchmarks and comments.

PR #490, opened 2019-06-14
feat: add support for enableABTest parameter

marchelbling:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes: not documenting for now as the Algolia documentation is not updated yet
| BC breaks?        | no     
| Related Issue     | No
| Need Doc update   | yes

## Describe your change

The Search API recently introduced the `enableABTest`parameter to disable the effect of an A/B test (mostly for internal bots or when testing queries in the Dashboard). 
This parameter is needed for the Query Suggestion feature. 

Adding a previously unsupported search parameter. Using URL parameters does not work otherwise as the Go client performs `POST` requests.


aseure:
As discussed with @marchelbling, there's probably solution than adding custom and undocumented functional options one by one. I'll leave this PR opened for now but we'll soon address this issue by offering a way to pass arbitrary option parameters.

aseure:
Closing in favor of #491

PR #491, opened 2019-06-17
Handle arbitrary search options

aseure:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | #490 
| Need Doc update   | no

## Describe your change

This PR introduces a new option under the `opt` package named `opt.ExtraOptions` which lets user pass extra custom options in search requests (those may be not already implemented options or unofficially documented options).

For instance, one can now perform the following `Search` call to pass the `enableABTest` parameter:

```go
res, err := index.Search(
	"query",
	opt.AttributesToRetrieve("objectID", "name"),
	opt.ExtraOptions(map[string]interface{}{"enableABTest": true}),
	opt.ExtraHeaders(map[string]string{"X-My-Header": "Hello World!"}),
)
```

Before, it was not possible to use unimplemented (because forgotten or undocumented) parameters.


algolia/internal/opt/extra_options.go
```diff
@@ -0,0 +1,28 @@
...
+// ExtractExtraOptions returns the first found ExtraOptionsOption from the
+// given variadic arguments or nil otherwise. If multiple options are found, the
+// inner maps are merged.
+func ExtractExtraOptions(opts ...interface{}) *opt.ExtraOptionsOption {
+	merged := make(map[string]interface{})
+
+	for _, o := range opts {
+		if v, ok := o.(*opt.ExtraOptionsOption); ok {
+			for key, value := range v.Get() {
+				merged[key] = value
```
marchelbling: what if `key` was already set? should we return an error?

algolia/search/requests_search.go
```diff
@@ -1,18 +1,79 @@
...
+	// with the known ones.
+	//
+	// To do that, we create a new type at runtime using the reflect
+	// package and populate it with:
+	//
+	//   1. the known fields: "query" + any other fields from
+	//      QueryParams if they are present
+	//   2. the unknown fields from the ExtraParams map
+	//
+	// Once the new type has been created, we will it with the
```
marchelbling:
```suggestion
	// Once the new type has been created, we will populate it with the
```

algolia/search/requests_search.go
```diff
@@ -1,18 +1,79 @@
...
+	// By there, because we do have extra search parameters, it
+	// is needed to flatten them in the same JSON payload, along
+	// with the known ones.
+	//
+	// To do that, we create a new type at runtime using the reflect
+	// package and populate it with:
+	//
+	//   1. the known fields: "query" + any other fields from
+	//      QueryParams if they are present
+	//   2. the unknown fields from the ExtraParams map
```
marchelbling: cf my remark regarding key conflict in the `merged` map; I think we should allow to override a param that was set explicitely i.e. 
```
client.Search(query, opt.Analytics(false), opt.ExtraOptions(map[string]interface{}{"analytics": true}))
``` 
should _not_ work in my opinion.

To extend on this, iirc, 
* `client.Search(query, opt.Analytics(false), opt.ExtraOptions(map[string]interface{}{"analytics": true}))` would end up with `analytics=true` 
* `client.Search(query, opt.ExtraOptions(map[string]interface{}{"analytics": true}), opt.Analytics(false))` would end up with `analytics=false`.

Note that I'm not asking to fail because the parameter has an explicit version but because there is a key conflict. Preventing parameters that can be explicitely set would be nice to have imo but maybe overkill.

algolia/internal/opt/extra_options.go
aseure: For other `opt.Extract*` functions in the client which operate on maps, we merge and override, this is the behavior we've choosen.

For instance, if you're doing:
```
index.Search(
	"",
	opt.ExtraHeaders(map[string]string{"My-Header": "Value 1"}),
	opt.ExtraHeaders(map[string]string{"My-Header": "Value 2"}),
)
```

You're gonna endup sending `Value 2` as an extra header.

**TL;DR** Consistency with other `Extract*` on maps.

algolia/search/requests_search.go
aseure: I think both point of views are valid actually. For the `opt.XXX > ExtraOptions(XXX)` that you are suggesting, I see your explanation as a _pro_ but `opt.ExtraURLParams` and `opt.ExtraHeaders` are actually overriding the values of any existing header. Not that those `opt.Extra` are advanced options that no one shoud use.

In short, while I see why your point is valid as wel, again for consistency with other `opt.Extra*` options, I would keep the following behavior: "if you are using `opt.Extra*` options, you know what you are doing, those are advanced options".

WDYT @Ant-hem

Ant-hem: I'd go for overriding the values with `opt` because overriding with `opt` is the current behavior for in the all the clients.

marchelbling: so if we have a consistent behavior across clients let's go for it. 
From a dev perspective, I would expect to have the extra params to typically be generated by code vs explicit options being "hardcoded" so I would find it safer to not have any overriding.

As Antoine said, we should have an explicit test on the overriding behavior!

algolia/internal/opt/extra_options.go
marchelbling: 👍

algolia/search/requests_search.go
marchelbling: Adn as a follow up, maybe it's already done and you can discard my comment, but this should be really explicitely documented.

PR #492, opened 2019-06-17
Add extra parameters

aseure:


Ant-hem:
@aseure  Will you open another PR for `userData` and `similarQuery` ?

aseure:
@Ant-hem As far as I can tell, `similarQuery` is only a search parameter and it is implemented in the first commit of this PR. So I don't think it needs to be implemented elsewhere (in settings for instance). However, for `userData` in the settings, yes, I'll do that in another PR (issue is still opened for that).

Ant-hem:
My bad regarding `similarQuery`. I read too fast.

PR #500, opened 2019-06-25
perf: benchmarks for getObjectID/hasObjectID functions

aseure:
**Summary**

As suggested by @pavel-popov on Github in
https://github.com/algolia/algoliasearch-client-go/pull/482
we should reconsider the implementation of `hasObjectID`/`getObjectID`
functions as the first implementation most probably behaves very poorly
in term of performances.

This commit contains 3 different implementations with their associated
tests and benchmarks to help us take a decision.

**Proposed implementations**

1. Marshaling-based implementation (original implemantion)

The original implementation relies on calls to
`json.Marshal`/`json.Unmarshal` to detect which fields is serialized to
`objectID` once encoded into JSON by the standard library. This method
has the advantage of respecting any custom `json.MarshalJSON` function
the user would have provided. However, we expect it to have poor
performances, each object needing to be encoded/decoded.

2. Interface-based implementation

As suggested by @pavel-popov in the previously mentioned Github PR, we
could ask users to implement a specific interface, such as:

```go
type ObjectIDProvider interface {
    ObjectID() string
}
```

This would ease the implementation of `getObjectID` where we would only
need an explicit cast from the `interface{}` to this custom interface
and a call to the `ObjectID()` method to retrieve the objectID. The
disadvantage of this method being that users would need to add
Algolia-related code (implementation of our interface) to improve
internal performances of our library.

3. Reflection-based implementation

The last implementation we suggest is a reflection-based one where the
objectID is retrieved from the `objectID` JSON tag, if any. By using
reflection, we would be able to inspect each object at runtime without
the need of Algolia-related code into the user-code (the use of JSON
tags is very Go idiomatic).

**Results**

```
BenchmarkHasObjectID/hasObjectID-8                300000      4791 ns/op
BenchmarkHasObjectID/hasObjectIDWithProvider-8  20000000      78.8 ns/op
BenchmarkHasObjectID/hasObjectIDWithReflect-8    5000000       316 ns/op
```

Results clearly shows that the interface-based implementation is the
fastest whereas the original implementation is confirmed to be the
slowest one (with a 61x slowdown factor).

Also, the reflection-based implementation is only 4x slower than the
interface-based one, while providing a 15x speedup compared our base
implementation.

**Suggestion**

Following what the results showed us, and as an effort to keep the
backward-compatibility with previous versions, I'd recommend to use the
reflection-based implementation as a first-intent in `getObjectID`
function to retrieve the value of the `objectID` JSON tag and then
fallback to the original implementation if it has not been found. This
way, we ensure that our implementation will continue to work with users
who implemented `json.MarshalJSON` on their Go structures while those
who directly used JSON tags will get the performance speedup for free
(i.e. without any implementation of Algolia-related interfaces).


aseure:
The first commit of this PR still contains the different implementations as well as their respective benchmarks for later reference. I've just added a second commit which reflects the decision we suggested (reflection for 1st intent, `json.Marshal`/`json.Unmarshal` for 2nd intent) to be ready when we will consider merging it.

Ant-hem:
Wouldn't it be nice to add some tests about this feature? I did [some in the Java client](https://github.com/algolia/algoliasearch-client-java-2/blob/febbaeafd7fbdee4affc9856a4601947a1bd1443/algoliasearch-core/src/test/java/com/algolia/search/EnsureObjectIDTest.java).

aseure:
@Ant-hem tests are already covered in [utils_test.go](https://github.com/algolia/algoliasearch-client-go/blob/5bdccf0e314fe806cd4b3e86948b79fad8bb7025/algolia/search/utils_test.go#L33-L55) in the same directory and they are still passing ✅

Ant-hem:
All good then!

PR #513, opened 2019-07-08
feat(index): add exists method

### LINKED ISSUE #512: feat: add index exists method
The spec and tests for this new method are described here: https://github.com/algolia/algoliasearch-client-specs/pull/26

sarahdayan:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | no     
| Related Issue     | Fix #512  
| Need Doc update   | yes

## Describe your change

This adds the `Index.Exists` method in the client. It follows the [spec](https://github.com/algolia/algoliasearch-client-specs/blob/master/features/index-exists.md) and [common test suite](https://github.com/algolia/algoliasearch-client-specs/tree/master/common-test-suite#Exists).

### DIFF
--- a/algolia/search/index.go
+++ b/algolia/search/index.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 
 	"github.com/algolia/algoliasearch-client-go/algolia/call"
+	"github.com/algolia/algoliasearch-client-go/algolia/errs"
 	iopt "github.com/algolia/algoliasearch-client-go/algolia/internal/opt"
 	"github.com/algolia/algoliasearch-client-go/algolia/transport"
 )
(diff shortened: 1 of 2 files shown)


ChangeLog.md
```diff
@@ -1,5 +1,11 @@
 # ChangeLog
 
+## [Unreleased](https://github.com/algolia/algoliasearch-client-go/compare/2.28.0...master) (2019-07-08)
+
+### Changes
+
+- **added:** Implement Index.Exists ([21b33e4](https://github.com/algolia/algoliasearch-client-go/commit/21b33e4))
```
aseure: Because we now generates the ChangeLog at release time using our internal tool `polyglot`, you should not need an extra commit to update the ChangeLog. Simply add a line at the end of your commit message with `[changelog]`. More details there: https://github.com/algolia/polyglot#polyglot-changelog. Note that this is not made explicit and public yet but we're planning on doing it soon on all API clients.

algolia/search/index.go
```diff
@@ -96,3 +97,10 @@ func (i *Index) GetStatus(taskID int) (res TaskStatusRes, err error) {
 	err = i.transport.Request(&res, http.MethodGet, path, nil, call.Read)
 	return
 }
+
+// Exists returns whether an initialized index exists or not.
+func (i *Index) Exists() bool {
+	_, err := i.GetSettings()
+	_, ok := errs.IsAlgoliaErrWithCode(err, http.StatusNotFound)
+	return !ok
```
aseure: Again, this implementation is just perfect if we want to consider network errors as `index does not exist` but according to what we discussed in person and what's already implemented in the PHP client: let's change a bit the signature here. We probably want to:

 1. have a return type of `(bool, error)` instead of just `(bool)`
 2. early return if the `GetSettings` returns a non-Algolia API error (you can check that thanks to the `errs.IsAlgoliaErr()` helper)

ChangeLog.md
sarahdayan: Reverted, and added the line in one of my commit.

algolia/search/index.go
sarahdayan: Done 👍

sarahdayan:
@aseure Everything is squashed into a single commit 🙂

PR #515, opened 2019-07-10
Gzip compression post put

aseure:


algolia/transport/transport.go
```diff
@@ -193,7 +224,21 @@ func buildRequest(
...
+		var r io.ReadCloser
+
+		if shouldCompress(c, method) {
+			switch c {
+			case compression.GZIP:
+				r = gzipCompressionReadCloser(data)
+			default:
+				r = noCompressionReadCloser(data)
+			}
+			headers["Content-Encoding"] = "gzip"
```
Ant-hem: Is the standard http client setting the `Content-Lenght` header as well?

Ant-hem:
Your CI tests are failing, you could disable the `GZIP` feature on the second test client to make the `accout_client` test passing.

algolia/transport/transport.go
aseure: Yes indeed. As per the documentation of `http.Request.Header`:
```
// For client requests, certain headers such as Content-Length
// and Connection are automatically written when needed and
// values in Header may be ignored. See the documentation
// for the Request.Write method.
```

aseure:
@Ant-hem I've modified some commits as I was globally modifying the default headers passed to every query :/ Now everything is fixed but the `AccountClient.CopyIndex` of course because one of the testing application is running on a cluster where the GZIP feature is not yet enabled. Waiting for this before merging. Feel free to approve once you're fine with the PR.

Issue #543, opened 2019-09-20
Readme instructions don't appear correct


ChrisMcKee:
Following the instructions in the readme verbaitm fails. 

Go 1.13 / ubuntu + windows.

```
$ go mod init playground
go: creating new go.mod: module playground

$ go get github.com/algolia/algoliasearch-client-go@v3.3.0

go: finding github.com v3.3.0
go: finding github.com/algolia/algoliasearch-client-go v3.3.0
go: finding github.com/algolia v3.3.0
go: finding github.com/algolia/algoliasearch-client-go v3.3.0
go get github.com/algolia/algoliasearch-client-go@v3.3.0: github.com/algolia/algoliasearch-client-go@v3.3.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v3
```

aseure:
Hello @ChrisMcKee, thank you very much for taking the time to report this issue.

The issue was already there before but cannot be mitigated anymore since Go 1.13 (I've been able to reproduce today with Go 1.13).

The origin of this issue is that when we introduced the use of Go modules in `v3.0.0`, we did not use the `v3/` prefix in module declarations. However, this seems mandatory for any published module with at least two major versions released [as documented in the golang wiki](https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher).

As for a solution, as of today, I cannot see any workaround that would not imply a breaking-change on the user-end. I've forked the repository to try to make a clean v3 release that would work with the `go get`/`go build` toolchain. On the user side, it means changing all import path from source files and the require statement of the `go.mod` file to append the needed `v3/` suffix. This can be done with a shell one-liner. We've been discussing it with the team internally and we are planning on introducing this breaking-change in a new minor version, along with the shell one-liner. I'm going to prepare a PR for that will wait for your feedback @ChrisMcKee.

cc @Ant-hem

ChrisMcKee:
Sounds fine; I'd read the issue in a few other places and happily I've been around long enough to know how to get around it locally using normal version referencing. Just a good one to get right in the first thing people come across (readme being reflected in the docs as well) 

I'll happily give it a test when its ready; as I say, not a show-stopper but incorrect / broken as far as instructions go 👍

aseure:
Closed by #544 and released as of [`v3.4.0`](https://github.com/algolia/algoliasearch-client-go/blob/master/ChangeLog.md#340-2019-09-25)

PR #544, opened 2019-09-24
fix: correctly migrate to Go modules by using v3/ suffix for Semantic Import Versioning

aseure:
[changelog]

**Manual Intervention Required**

Our migration to Go modules starting at version v3.0.0 of this Go API
client was done incorrectly since our module declaration was not
including the required `v3/` suffix. This commit fixes this issue by
respecting the Semantic Import Versioning described by the Go wiki.

**Step 1/2**

To update your project which depends on the Algolia Go API client, you
need to replace all `github.com/algolia/algoliasearch-client-go/*`
import statements with `github.com/algolia/algoliasearch-client-go/v3/*`
instead. The following shell one-liner can be used to perform this
change:

```
for f in $(find . -type f); do
  sed -i '' 's:github.com/algolia/algoliasearch-client-go:github.com/algolia/algoliasearch-client-go/v3:g' $f
done

for f in $(find . -type f); do
  sed -i 's:github.com/algolia/algoliasearch-client-go:github.com/algolia/algoliasearch-client-go/v3:g' $f
done
```

**Step 2/2**

After that, make sure to import the Algolia Go client dependency as such
in your `go.mod` file:

```
require github.com/algolia/algoliasearch-client-go/v3 v3.X.Y
```

where `v3.X.z` stands for the exact release tag you want to use.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,9 @@ os:
   - linux
 
 go:
-  - "1.9"
+  - "1.11"
   - "1.12"
-
-before_install:
-  - go get -u github.com/rakyll/gotest
-  - curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+  - "1.13"
 
 install:
   - make deps
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 all: deps generate lint test
 
 deps:
-	dep ensure -vendor-only
 	go get ./...
 
 generate:
@@ -13,10 +12,10 @@ lint:
 test: unit-tests integration-tests
 
 unit-tests:
-	gotest -v ./algolia/...
+	go test -v ./algolia/...
 
 integration-tests:
-	gotest -v ./cts/...
+	go test -v ./cts/...
 
 clean:
 	rm -f `grep -R -l --include \*.go -F 'DO NOT EDIT' * | grep -v -F 'algolia/internal/gen/'`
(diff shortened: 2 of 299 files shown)


aseure:
Fix #543 

@ChrisMcKee @Ant-hem 
I've tagged this commit with `v3.4.0-beta.1`, could you try the following and tell me if this works for you with Go 1.13:

 1. update an existing Go project of yours (or create a new one) with the correct `github.com/algolia/algoliasearch-client-go/v3/*` import statements
 2. run `go get github.com/algolia/algoliasearch-client-go/v3@v3.4.0-beta.1`
 3. run `go build`

I've tried it on my end and it seems to work perfectly now. For the record, here are my project files:

**main.go**
```go
package main

import "github.com/algolia/algoliasearch-client-go/v3/algolia/search"

func main() {
	c := search.NewClient("", "")
	_ = c
}
```

**go.mod**
```
module tmp

go 1.13

require github.com/algolia/algoliasearch-client-go/v3 v3.4.0-beta.1
```

ChrisMcKee:
Cool i'll give it a go. One tick

ChrisMcKee:
image

image

Looks good

aseure:
Released as of [`v3.4.0`](https://github.com/algolia/algoliasearch-client-go/blob/master/ChangeLog.md#340-2019-09-25)

Thank you again @ChrisMcKee and @Ant-hem for double-checking this solution. Sadly, this will imply some manual changes on the user side for users who already have a v3 codebase but it's better for the long term to get this module story right. I've also updated the documentation snippets internally at Algolia, it should be reflected today in our documentation at https://www.algolia.com/doc.

PR #546, opened 2019-09-25
feat(key): add Client.GetSecuredAPIKeyRemainingValidity() (#536)

### LINKED ISSUE #536: New method: GetSecuredApiKeyRemainingValidity
## Summary

When a secured API keys expire, search breaks. The current workaround is more complex
than it needs to be. 

The options for developers using Algolia require, at every place where a search
operation is performed, catching an operation error, identify it as a
"ValidUntil" error, then regenerate the API key, and perform the search
operation again.

The idea is to define a method which tells if a `SecuredAPIKey`
is expired or not without making a single call to Algolia's API.

## References

- Specs: https://github.com/algolia/algoliasearch-client-specs-internal/pull/25

aseure:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fix #536
| Need Doc update   | yes

## Describe your change

The remaining validity duration of a secured API generated with the
`opt.ValidUntil()` parameter can now be checked thanks to the new
`GetSecuredAPIKeyRemainingValidity()` method of the `search.Client`.

### DIFF
--- a/algolia/errs/errors.go
+++ b/algolia/errs/errors.go
@@ -6,12 +6,16 @@ import (
 )
 
 var (
-	ErrMissingObjectID    = errors.New("objectID is missing or empty")
-	ErrMissingKeyID       = errors.New("key ID is not set in Key.Value")
-	ErrNoMoreHostToTry    = errors.New("all hosts have been contacted unsuccessfully, it can either be a network error or wrong appID/key credentials were used")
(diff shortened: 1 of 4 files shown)


aseure:
@Ant-hem Fixed. Could you approve if everything's good on your side?

PR #548, opened 2019-09-25
Chore/lint on travis build

aseure:


algolia/transport/transport.go
```diff
@@ -178,26 +178,6 @@ func mergeHeaders(defaultHeaders, extraHeaders map[string]string) map[string]str
 	return headers
 }
 
-func gzipCompressionReadCloser(in *io.PipeReader) io.ReadCloser {
```
Ant-hem: Why is this one removed? Not used anymore?

aseure: Yes, dead code that was still there. The new and used function is just below, it is named `wrapGZIPEncoder`.

PR #550, opened 2019-10-02
Invalid validuntil for get secured api key remaining validity

aseure:
**Summary**

The `errs.ErrValidUntilInvalid` error was introduced to signify that the
`validUntil` string parameter of a secured API key could not be parsed
as an integer by the `Client.GetSecuredAPIKeyRemainingValidity()`
method. However, since this string value is always extracted thanks to a
regexp looking for a integer of maximum length 10, the conversion can
never fail, making this error case useless.

This commit hence removes this useless error which get never returned.

### DIFF
--- a/cts/search/secured_api_keys_test.go
+++ b/cts/search/secured_api_keys_test.go
@@ -108,4 +108,12 @@ func TestSecuredAPIKeyParametersValidity(t *testing.T) {
 	require.NoError(t, err)
 	_, err = c.GetSecuredAPIKeyRemainingValidity(securedKey)
 	require.EqualError(t, err, errs.ErrValidUntilNotFound.Error())
+
+	securedKey = cts.GenerateSecuredAPIKeyWithArbitraryParameters(
+		t,
+		searchKey,
+		map[string]interface{}{"validUntil": "NaN"},
+	)
+	_, err = c.GetSecuredAPIKeyRemainingValidity(securedKey)
+	require.EqualError(t, err, errs.ErrValidUntilNotFound.Error())
 }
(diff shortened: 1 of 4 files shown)


cts/search/secured_api_keys_test.go
```diff
@@ -108,4 +108,12 @@ func TestSecuredAPIKeyParametersValidity(t *testing.T) {
...
 	_, err = c.GetSecuredAPIKeyRemainingValidity(securedKey)
 	require.EqualError(t, err, errs.ErrValidUntilNotFound.Error())
+
+	securedKey = cts.GenerateSecuredAPIKeyWithArbitraryParameters(
+		t,
+		searchKey,
+		map[string]interface{}{"validUntil": "NaN"},
+	)
+	_, err = c.GetSecuredAPIKeyRemainingValidity(securedKey)
+	require.EqualError(t, err, errs.ErrValidUntilNotFound.Error())
```
samouss: Maybe the error could be renamed to match broader use cases? The function does return an error when it's not found but also when it's invalid. From the user perspective it might be confusing to get such error when the provided value contains `validUntil` but invalid.

aseure: `errs.ErrNoIntegerValidUntilFound` would be the closest thing we can do. Since we find this value thanks to a regexp, not finding an integer value for `validUntil` and not finding `validUntil` at all is the same. WDYT? @samouss

samouss: We could keep the same error name but use a different error message. Indeed you're correct they are almost the same, in the end it's not found. Maybe with an error message like: "no valid `validUntil` parameter found, please make sure the secured API key has been generated correctly". It's a bit redundant but you got the idea. WDYT?

Ant-hem: Yes! I like this message - it's explicit and it's almost what we are sending in both Java/C# client.

```java
"The Secured API Key doesn't have a validUntil parameter."
```

https://github.com/algolia/algoliasearch-client-java-2/blob/6e3cf8543d3773e6b53b6316ba8884781c2200ff/algoliasearch-core/src/main/java/com/algolia/search/SearchClientAPIKeys.java#L451

aseure: Great, I'll change the description to this one then. Thank you both.

aseure:
@Ant-hem @samouss Same PR, I've simply rebased and run the missing lint, could you please re-approve. Thanks!

PR #553, opened 2019-10-04
Contextual Query Rules

### LINKED ISSUE #542: feat: add contextual rules
**TL;DR**: The `condition` field on rules is now optional.

Actions required:

- Update the `cts` with the information on : [#71](https://github.com/algolia/algoliasearch-client-specs-internal/pull/71/files)
- On strongly typed clients, update types in query rules, as the `condition` field in rules is now **optional**

- [Python test suite update](https://github.com/algolia/algoliasearch-client-python/pull/450/files)

aseure:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fix #542 
| Need Doc update   | no

## Describe your change

Makes Query Rules' condition and condition fields optional to let query rules trigger contextually (when using `opt.RuleContexts` at search time).

@Ant-hem @nunomaduro Skip the review of the first commit, this is a bug fix on generated files (a better null-checking). The contextual query rules related change is fairly small.

### DIFF
--- a/algolia/search/rule.go
+++ b/algolia/search/rule.go
@@ -66,6 +66,22 @@ func (r Rule) Equal(r2 Rule) bool {
 	return true
 }
 
+func (r Rule) MarshalJSON() ([]byte, error) {
+	type _Rule Rule
+	type _RuleWithOptionalCondition struct {
+		Condition *RuleCondition `json:"condition,omitempty"`
+		_Rule
+	}
+
+	var tmp _RuleWithOptionalCondition
+	tmp._Rule = _Rule(r)
+
+	if r.Condition != (RuleCondition{}) {
+		tmp.Condition = &r.Condition
+	}
+	return json.Marshal(tmp)
+}
+
 // Equal returns true if the TimeRanges are equal. It returns false otherwise.
 func (r TimeRange) Equal(r2 TimeRange) bool {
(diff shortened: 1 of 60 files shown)


algolia/search/rule.go
```diff
@@ -66,6 +66,22 @@ func (r Rule) Equal(r2 Rule) bool {
 	return true
 }
 
+func (r Rule) MarshalJSON() ([]byte, error) {
+	type _Rule Rule
+	type _RuleWithOptionalCondition struct {
+		Condition *RuleCondition `json:"condition,omitempty"`
```
Ant-hem: I might be missing something here, but why don't we set `json:"condition,omitempty"` in the `Rule struct` directly? Instead of creating a tmp one.

https://github.com/algolia/algoliasearch-client-go/blob/9d7e6def70a5ea50700562d7651e7b572192eed6/algolia/search/rule.go#L14

aseure: Sadly, `omitempty` only works on literal types, not user-defined structures. From the [Go documentation](https://golang.org/pkg/encoding/json/#Marshal):

> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.

Ant-hem: Oh too bad :(
All good then.

PR #555, opened 2019-10-09
Fix aa testing test

aseure:


aseure:
@Ant-hem Because the two PRs were stacked, I had to re-push this one, could you re-approve (nothing changed), sorry about the noise.

PR #566, opened 2019-11-17
chore: remove dep related dependency files

ferhatelmas:
## Describe your change

Drop dep files since go modules are leveraged.

Code cleanup.


aseure:
Thank you for your contribution @ferhatelmas!

Indeed, the Go client is now fully compatible with Go modules. We've kept the `dep`-related files to make sure the project was still fetchable by `dep` users. I'm not yet sure of the impact of removing those files for `dep` users. If you already have the answer to this question, feel free to redirect us to the right source of documentation. Otherwise, I'll try to find some time next week to test with `dep` and with Go modules disabled.

aseure:
Hello @ferhatelmas and happy new year 🎉 

I found the time to try the approach of removing `dep`-related files from a Go library and I could not find any issue that could occur for remaining `dep` users since the project is following semver.

Hence, we'll be able to merge your request. Thank you for taking the time to contribute.

PR #570, opened 2019-12-13
feat(transport): expose the default HTTP client via transport.DefaultHTTClient()

aseure:
transport.DefaultHTTPClient exposes the default *http.Client used by the
different Client instances of the Algolia API client.

Most users should not need to access this http.Client.

This helper is only useful for users who wish to keep the default behavior
of the Algolia API client (by wrapping it in a user-defined instance of
transport.Requester) but would like to wrap it into a middleware layer or pass
it to an HTTP interceptor.

### DIFF
--- a/algolia/transport/requester.go
+++ b/algolia/transport/requester.go
@@ -13,6 +13,30 @@ const (
 	DefaultTLSHandshakeTimeout = 2 * time.Second
 )
 
+// DefaultHTTPClient exposes the default *http.Client used by the different
+// Client instances of the Algolia API client.
+//
+// Most users should not need to access this http.Client.
+//
+// This helper is only useful for users who wish to keep the default behavior of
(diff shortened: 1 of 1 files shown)


aseure:
> Is there a specific reason not to re-use the same client for different requesters?

We wanted to keep all the following functions "standalone" by default:

 - `search.NewClient()`
 - `analytics.NewClient()`
 - `insights.NewClient()`

We had no idea on how to keep it simple while reusing the same underlying HTTP client and without opening up too much the public API. Also, we had in mind to use dedicated HTTP client instances if one day we need to target the different Algolia REST APIs with different settings.

I recognize that the current PR will indeed expose the default HTTP client through a function which could be end up being deprecated if we start to have dedicated HTTP client for each `{search,analytics,insights}.Client` but I think it's an acceptable trade off.

PR #573, opened 2020-01-08
fix(rule): correctly serialize RuleCondition

### LINKED ISSUE #572: Not possible to create Query Rule triggered only by empty query
- Algolia Client Version: 3.5.0
- Language Version: go1.13.3 darwin/amd64

### Description

Hello team! 👋 

I need to create query rule for an empty query with Go API client. However, I’m getting `Algolia API error [400] Missing mandatory pattern (condition.pattern) (in [88]) (near 1:68469)` from the API. 

I believe it’s due to [`json:"pattern,*omitempty*"`](https://github.com/algolia/algoliasearch-client-go/blob/master/algolia/search/rule_condition.go#L7), which removes the pattern completely from generated JSON and therefore API rejects the call, and was introduced by https://github.com/algolia/algoliasearch-client-go/commit/f2cdcdd582adaa58f4ad50edd449485a5b390229 in 3.5.0 version of the client.

In [documentation](https://www.algolia.com/doc/api-reference/api-methods/save-rule/#method-param-pattern) I found that `""` pattern should be allowed:

> This parameter goes hand in hand with the anchoring parameter. If you're creating a Rule that depends on a specific query, you must specify the pattern and anchoring. The empty "" pattern is only allowed when anchoring is set to is.

However it doesn’t work for me and I didn’t manage to find the corresponding code, which should handle the pattern based on `anchoring.Is` 😕 

### Steps To Reproduce
Try to push query rule with condition:
```go
Condition: search.RuleCondition{
	Anchoring: search.Is,
	Pattern:   "",
},
```

Thanks for taking a look!

aseure:
Since the introduction of Contextual Query Rules in `f2cdcdd`, the
anchoring and pattern parts were omitted when serialized if one of them
was missing. However, the correct behavior is to serialized both fields
when at least of them is present. This commit fixes this.

Close #572

### DIFF
--- a/algolia/search/rule_condition.go
+++ b/algolia/search/rule_condition.go
@@ -3,10 +3,25 @@ package search
 import "encoding/json"
 
 type RuleCondition struct {
-	Anchoring    RulePatternAnchoring `json:"anchoring,omitempty"`
-	Pattern      string               `json:"pattern,omitempty"`
-	Context      string               `json:"context,omitempty"`
-	Alternatives *Alternatives        `json:"alternatives,omitempty"`
+	Anchoring    RulePatternAnchoring
+	Pattern      string
+	Context      string
+	Alternatives *Alternatives
+}
+
+func (c RuleCondition) MarshalJSON() ([]byte, error) {
+	m := make(map[string]interface{})
+	if c.Anchoring != "" || c.Pattern != "" {
+		m["anchoring"] = c.Anchoring
+		m["pattern"] = c.Pattern
+	}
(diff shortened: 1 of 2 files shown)


JanPetr:
Looks good to me! 👍 I'll let @Ant-hem to approve as I don't have a deep knowledge in the API clients.

JanPetr:
Hello @aseure, @Ant-hem 👋  How does it look with this PR? When can we expect that to be part of the API client?

PR #575, opened 2020-01-10
fix(search): correctly encode index name in URIs (#574)

### LINKED ISSUE #574: Wrongly generated URL when hashtag in index name
* Algolia for Go (3.5.0);
* Go (go1.13.4)

### Description
It seems like additional parameters like `getVersion=2` get interleaved with the index name when it contains a hashtag.
I suppose the hashtag needs to be encoded so that it is not considered as an actual hashtag when leveraging URL helpers.

### How ro reproduce
```go
package main

import (
	"log"

	"github.com/algolia/algoliasearch-client-go/v3/algolia/debug"
)

func main() {
	debug.Enable()
	c := search.NewClient("YYY", "XXX")
	i := c.InitIndex("indexname@algolia.com#whatever")
	s, err := i.GetSettings()
	if err != nil {
		log.Printf("err: %v", err)
		return
	}

	log.Printf("settings: %v", s)
}

```

outputs:

```
ALGOLIA DEBUG: > ALGOLIA DEBUG request:
	method="GET"
	url="https://YYY-dsn.algolia.net/1/indexes/indexname@algolia.com?getVersion=2#whatever/settings"
	header=User-Agent:"Algolia for Go (3.5.0);Go (go1.13.4)"
	header=X-Algolia-Application-Id:"**********"
	header=X-Algolia-Api-Key:"********************************"
	header=Connection:"Keep-Alive"
	header=Content-Type:"application/json; charset=utf-8"
	body=
```

### How this was discovered
Using an API key with restriction strictly on the full index name made the above request fail because it would target `indexname@algolia.com` instead of `indexname@algolia.com#whatever`.

aseure:
We've recently identified that the user-provided index name, when passed
to the URL internally, was not URL-encoded as one would expect. This
could result in the user targetting wrong indices, either when using the Go
client, or other API clients, including frontend integrations.

After multiple internal discussions, we finally decided to introduce
this change, even though it may be breaking for some users, in a minor
version. We consider that the future benefits of fixing this issue for
future new users holds more positive impact than the negative one of
having existing users to have to manually ensure their index names are
correct.

Close #574

### DIFF
--- a/algolia/search/index.go
+++ b/algolia/search/index.go
@@ -3,6 +3,7 @@ package search
 import (
 	"fmt"
 	"net/http"
+	"net/url"
 
 	"github.com/algolia/algoliasearch-client-go/v3/algolia/call"
 	"github.com/algolia/algoliasearch-client-go/v3/algolia/errs"
@@ -31,7 +32,7 @@ func newIndex(client *Client, name string) *Index {
 }
 
 func (i *Index) path(format string, a ...interface{}) string {
-	prefix := fmt.Sprintf("/1/indexes/%s", i.name)
+	prefix := fmt.Sprintf("/1/indexes/%s", url.QueryEscape(i.name))
 	suffix := fmt.Sprintf(format, a...)
 	return prefix + suffix
 }
--- a/cts/index/url_encoding_index_name_test.go
+++ b/cts/index/url_encoding_index_name_test.go
@@ -1,0 +1,67 @@
+package index
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+	"unicode"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/algolia/algoliasearch-client-go/v3/algolia/wait"
+	"github.com/algolia/algoliasearch-client-go/v3/cts"
+)
+
+func TestEnableURLEncodingIndexName(t *testing.T) {
+	t.Parallel()
+
+	var (
+		client              = cts.InitSearchClient1(t)
+		indexNames          []string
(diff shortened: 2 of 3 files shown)


algolia/search/index.go
```diff
@@ -31,7 +32,7 @@ func newIndex(client *Client, name string) *Index {
 }
 
 func (i *Index) path(format string, a ...interface{}) string {
-	prefix := fmt.Sprintf("/1/indexes/%s", i.name)
+	prefix := fmt.Sprintf("/1/indexes/%s", url.QueryEscape(i.name))
```
nunomaduro: I am afraid that this may be breaking change for existing customers. 🙁

rayrutjes: Definitely a breaking change, but a bug fix so there is no way around it.
Currently the behavior is very dangerous as it transforms index names on the fly by truncating starting from the hashtag.

aseure: Agree with @rayrutjes on this. Changing it is bad for some users on the short-term but not doing anything is bad for all users on the long term. I'm in favor of changing this even if it means introducing a breaking change. I'll check with @robertmogos on the feasability of emailing all Go client users to let them know about the bug and the bug fix. In any case, this will end up in the ChangeLog.

nunomaduro: Good call @aseure!

rayrutjes: I'd actually call this a fixing change if that can make everyone feel better :)

aseure:
@Ant-hem The fix has been confirmed by the rest of the team. @Ant-hem Could you review the code itself and let me know if you see anything wrong. Thanks!

PR #577, opened 2020-01-24
fix: implement correct type for `Settings.UserData`

aseure:
So far, the inner value of `opt.UserDataOption` was incorrectly typed as a
`map[string]interface{}` while the engine can actually accepts both JSON
objects or JSON literals.

This commit ensures that any literal can now be accepted as a valid input of
`opt.UserData(...)` (by accepting an `interface{}` instead of a map).

To also be able to correctly decode the inner value of the
`opt.UserDataOption`, the method `opt.UserData.Decode(..)` was also
added to let the users decode the inner value into any Go variable.


aseure:
More user-oriented point of view, this lets us do:

```go
var yourVariable string // or really any type
err := settings.UserData.Decode(&yourVariable)
```

where before we could only do

```go
var m map[string]interface{}
m = settings.UserData.Get() // panics if the `userData` is not a JSON object
```

PR #590, opened 2020-03-13
fix(analytics): accept region.DE instead of region.EU for Analytics configuration

aseure:
The `region.EU` was never accepted by the Analytics API. This commit
introduces the right region `region.DE` and uses it as part of the valid
configurable regions for the Analytics client.

### DIFF
--- a/algolia/analytics/utils.go
+++ b/algolia/analytics/utils.go
@@ -10,7 +10,7 @@ import (
 
 func defaultHosts(r region.Region) (hosts []*transport.StatefulHost) {
 	switch r {
-	case region.EU, region.US:
+	case region.DE, region.US:
 		hosts = append(hosts, transport.NewStatefulHost(fmt.Sprintf("analytics.%s.algolia.com", r), call.IsReadWrite))
 	default:
 		hosts = append(hosts, transport.NewStatefulHost("analytics.algolia.com", call.IsReadWrite))
--- a/algolia/region/region.go
+++ b/algolia/region/region.go
@@ -5,4 +5,5 @@ type Region string
 const (
 	US Region = "us"
 	EU Region = "eu"
+	DE Region = "de"
 )


aseure:
Thank you again for noticing and reporting @nunomaduro 🙏

algolia/analytics/utils.go
```diff
@@ -10,7 +10,7 @@ import (
 
 func defaultHosts(r region.Region) (hosts []*transport.StatefulHost) {
 	switch r {
-	case region.EU, region.US:
+	case region.DE, region.US:
```
nunomaduro: This is a breaking change buddy, as people who here using `region.EU` will now be targetting `US`. I think this line should be:

```suggestion
	case region.DE, region.EU, region.US:
```

algolia/region/region.go
```diff
@@ -5,4 +5,5 @@ type Region string
 const (
 	US Region = "us"
 	EU Region = "eu"
```
Ant-hem: Looks like region `eu` doesn`t exist:

From https://www.algolia.com/doc/rest-api/analytics/
```
The following domains are available:

    United States: https://analytics.us.algolia.com
    Europe (Germany): https://analytics.de.algolia.com
```

nunomaduro: Yeah, that was the problem. That's why we are introducing this change. I think @aseure is leaving that enumeration to not break peoples code.

Ant-hem: It's fine to delete it IMO. No one should be using it as it doesn't exist right?

nunomaduro: @Ant-hem being pragmatic, yes. I agree. Let's delete it. What do you think @aseure ?

aseure: I'll have to double check that. I think it existed in the past but it was not documented and it triggered the DE cluster. I'll double check with the Analytics team.

aseure: I just confirmed with @marchelbling and we do have a `recommendation.eu.algolia.com` which prevents us from removing it. And for the record @sfriquet confirmed that the the `eu` region for the analytics endpoint never existed.

@Ant-hem in other words, we'll need to keep it since customers may be using it to instanciate the recommendation client.

PR #605, opened 2020-05-04
refactor(transport): ensure early termination upon context cancellation

### LINKED ISSUE #596: Client enters retry mode even if the parent context's deadline is exceeded causing all attempts to fail.
- Algolia Client Version: 3.5.2
- Language Version: go

### Description

If the parent context exceeds its deadline, the algolia client goes into a retry mode. However, all subsequent attempt will fail due to the parent context.

```
package main

import (
	"context"
	"fmt"
	"net/http"
	"time"
)

type Contact struct {
	ObjectID string `json:"objectID"`
	Name     string `json:"name"`
}

func main() {
	scenario("normal", 1*time.Second, 1*time.Second, false)
	scenario("short context", 1*time.Second, 1*time.Millisecond, false)
	scenario("short context & short timeout", 1*time.Millisecond, 1*time.Millisecond, false)
	scenario("early cancellation", 1*time.Second, 1*time.Second, true)
}

type requester struct {
	name string
}

func (r *requester) Request(req *http.Request) (*http.Response, error) {
	fmt.Println(r.name, req.URL)
	res, err := http.DefaultClient.Do(req)
	if res != nil {
		fmt.Println(r.name, res.StatusCode)
	}
	if err != nil {
		fmt.Println(r.name, err)
	}
	return res, err
}

func scenario(name string, writeTimeout time.Duration, contextTimeout time.Duration, cancelEarly bool) () {

	client := search.NewClientWithConfig(search.Configuration{
		APIKey:       "your_api_key",
		AppID:        "your_app_id",
		WriteTimeout: writeTimeout,
		Requester:    &requester{name: name},
	})
	index := client.InitIndex("your_index_name")

	// =======================
	// Using a short context timeout to exacerbate the issue
	ctx, cancel := context.WithTimeout(context.Background(), contextTimeout)
	defer cancel()

	if cancelEarly {
		cancel()
	}

	// Any function that accepts a context.
	res, err := index.SaveObjects([]Contact{
		{ObjectID: "1", Name: "Foo"},
	}, ctx, )

	fmt.Println(name, res, err)
}

```

This generates logs similar to
```
normal https://your_app_id.algolia.net/1/indexes/your_index_name/batch
normal 200
normal {[{[1] 493632544001 0x1272ea0}]} <nil>
short context https://your_app_id.algolia.net/1/indexes/your_index_name/batch
short context Post https://your_app_id.algolia.net/1/indexes/your_index_name/batch: context deadline exceeded
short context https://your_app_id-2.algolianet.com/1/indexes/your_index_name/batch
short context Post https://your_app_id-2.algolianet.com/1/indexes/your_index_name/batch: context deadline exceeded
short context https://your_app_id-1.algolianet.com/1/indexes/your_index_name/batch
short context Post https://your_app_id-1.algolianet.com/1/indexes/your_index_name/batch: context deadline exceeded
short context https://your_app_id-3.algolianet.com/1/indexes/your_index_name/batch
short context Post https://your_app_id-3.algolianet.com/1/indexes/your_index_name/batch: context deadline exceeded
short context {[]} could not send intermediate batch: all hosts have been contacted unsuccessfully, it can either be a network error or wrong appID/key credentials were used
short context & short timeout https://your_app_id.algolia.net/1/indexes/your_index_name/batch
short context & short timeout Post https://your_app_id.algolia.net/1/indexes/your_index_name/batch: context deadline exceeded
short context & short timeout https://your_app_id-2.algolianet.com/1/indexes/your_index_name/batch
short context & short timeout Post https://your_app_id-2.algolianet.com/1/indexes/your_index_name/batch: context deadline exceeded
short context & short timeout https://your_app_id-3.algolianet.com/1/indexes/your_index_name/batch
short context & short timeout Post https://your_app_id-3.algolianet.com/1/indexes/your_index_name/batch: context deadline exceeded
short context & short timeout https://your_app_id-1.algolianet.com/1/indexes/your_index_name/batch
short context & short timeout Post https://your_app_id-1.algolianet.com/1/indexes/your_index_name/batch: context deadline exceeded
short context & short timeout {[]} could not send intermediate batch: all hosts have been contacted unsuccessfully, it can either be a network error or wrong appID/key credentials were used
early cancellation https://your_app_id.algolia.net/1/indexes/your_index_name/batch
early cancellation Post https://your_app_id.algolia.net/1/indexes/your_index_name/batch: context canceled
early cancellation https://your_app_id-1.algolianet.com/1/indexes/your_index_name/batch
early cancellation Post https://your_app_id-1.algolianet.com/1/indexes/your_index_name/batch: context canceled
early cancellation https://your_app_id-3.algolianet.com/1/indexes/your_index_name/batch
early cancellation Post https://your_app_id-3.algolianet.com/1/indexes/your_index_name/batch: context canceled
early cancellation https://your_app_id-2.algolianet.com/1/indexes/your_index_name/batch
early cancellation Post https://your_app_id-2.algolianet.com/1/indexes/your_index_name/batch: context canceled
early cancellation {[]} could not send intermediate batch: all hosts have been contacted unsuccessfully, it can either be a network error or wrong appID/key credentials were used
```

I would not expect all hosts to be contacted. Expectation is to exit as early as the context is done to avoid waste work and marking hosts as "timeout".
I believe that the issue comes from this code https://github.com/algolia/algoliasearch-client-go/blob/master/algolia/transport/transport.go#L115-L119.
Ideally, you would check the parent context `Done()` before retrying.

### Steps To Reproduce

Edit the code provided above with your app id, api key and a test index name.

aseure:
Because we were not checking if the error returning by the underlying
requester, we could not determine if the HTTP call failed because of
user-defined context being cancelled because of a timeout being reached or
an exceeding deadline.

This commit introduce an early return for those cases. This way, we do
not perform multiple retries against other hosts when the user-defined
context is already done, because of a context timeout or a context deadline.

Note that we did not use the error wrapping mechanism from Go 1.13 with
`errors.Is()` and `errors.As()` since we are still supporting Go down to
Go 1.11 for now.

Close #596

### DIFF
--- a/algolia/transport/transport.go
+++ b/algolia/transport/transport.go
@@ -116,6 +116,15 @@ func (t *Transport) Request(
 		req = req.WithContext(perRequestCtx)
 		bodyRes, code, err := t.request(req)
 
+		// Context error only returns a non-nil error upon context
+		// cancellation, which is a signal we interpret as an early return.
+		// Indeed, we do not want to retry on other hosts if the context is
+		// already cancelled.
+		if ctx.Err() != nil {
+			cancel()
+			return err
+		}
+
 		switch t.retryStrategy.Decide(h, code, err) {
 		case Success:
 			err = unmarshalTo(bodyRes, &res)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,13 +38,27 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-            key: go-cache-1-14
+          key: go-cache-1-14
       - run:
           name: Install golangci-lint linter
           command: curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.25.0
       - run:
           name: Check formatting
           command: make lint
+  test-1-11:
+    docker:
+      - image: circleci/golang:1.11
+    steps:
+      - checkout
+      - restore_cache:
+          key: go-cache-1-11
+      - credentials
+      - test:
+          go_version: "1.11"
+      - save_cache:
+          key: go-cache-1-11
+          paths:
+            - "~/go/pkg"
   test-1-12:
     docker:
       - image: circleci/golang:1.12
@@ -93,6 +107,7 @@ workflows:
   build:
     jobs:
       - format
+      - test-1-11
       - test-1-12
       - test-1-13
       - test-1-14
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,7 +12,6 @@ linters:
     - gofmt
     - goimports
     - golint
-    - gosimple
     - govet
     - ineffassign
     - misspell
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
     <a href="https://godoc.org/github.com/algolia/algoliasearch-client-go"><img src="https://godoc.org/github.com/algolia/algoliasearch-client-go?status.svg" alt="GoDoc"></img></a>
     <a href="https://goreportcard.com/report/github.com/algolia/algoliasearch-client-go"><img src="https://goreportcard.com/badge/github.com/algolia/algoliasearch-client-go" alt="Go Report Card"></img></a>
     <a href="https://github.com/algolia/algoliasearch-client-go/blob/master/LICENSE"><img src="https://img.shields.io/badge/license-MIT-blue.svg" alt="License"></img></a>
-    <img src="https://img.shields.io/badge/Go-%3E=1.8-green.svg" alt="Supported version"></img></a>
+    <img src="https://img.shields.io/badge/Go-%3E=1.11-green.svg" alt="Supported version"></img></a>
   </p>
 </p>
 
@@ -25,7 +25,7 @@
 
 ## ✨ Features
 
-* Support Go 1.8 and above
+* Support Go 1.11 and above
 * Typed requests and responses
 * First-class support for user-defined structures
 * Injectable HTTP client


algolia/transport/transport.go
```diff
@@ -140,29 +144,56 @@ func (t *Transport) Request(
...
+		isDeadlineExceeded = strings.Contains(err.Error(), context.DeadlineExceeded.Error())
+		isContextDone      bool
+	)
+
+	select {
+	case <-ctx.Done():
+		isContextDone = true
+	default:
+		isContextDone = false
+	}
```
samouss, line 163: `ctx.Err()` returns a `nil` error when it's not closed and a non-`nil` error when closed. Wouldn't be simpler to use the function exposed on the context rather than the select? I know it won't block but I struggle a bit to understand why we had a select in the first place.

```
// If Done is not yet closed, Err returns nil.
// If Done is closed, Err returns a non-nil error explaining why:
[...]
```

algolia/transport/transport.go
```diff
@@ -140,29 +144,56 @@ func (t *Transport) Request(
...
 }
 
+func isErrBecauseOfContextCancellation(ctx context.Context, err error) bool {
+	if err == nil {
+		return false
+	}
+
+	var (
+		isCancelled        = strings.Contains(err.Error(), context.Canceled.Error())
+		isDeadlineExceeded = strings.Contains(err.Error(), context.DeadlineExceeded.Error())
```
samouss, line 154: Why do we have to check the `err` value? Wouldn't be enough to check for the context? Whatever the error we got we probably want to stop if the context is canceled no?

aseure: Actually, I've forgot to take a step back from the original issue. Thanks for noticing. It makes sense to abort on any context cancellation, since the end goal is just to early abort the retry strategy when user-provided context is already cancelled.

Thanks for noticing!

algolia/transport/transport.go
aseure: Did not think about using the `.Err()` return value, thanks for noticing. I'll change that.

aseure:
Thanks for the review @samouss, would you mind approving as well after my last amend?

PR #611, opened 2020-06-03
feat(search): expose built-in operations for partial updates with PartialUpdateOperation

aseure:


algolia/search/requests_indexing.go
```diff
@@ -42,3 +42,8 @@ type getObjectsReq struct {
 type batchReq struct {
 	Requests []BatchOperation `json:"requests,omitempty"`
 }
+
+type PartialUpdateOperation struct {
```
nunomaduro, line 46: Can you explain the context buddy?

aseure: This related to the documentation change which requires to add snippets for partial update operations, such as, in Go:

```go
    type Record struct {
        ObjectID string                        `json:"objectID"`
        Count    search.PartialUpdateOperation `json:"count"`
    }
    res, err := index.PartialUpdateObjects([]Record{
        {
            ObjectID: "myID",
            Count: search.PartialUpdateOperation{
                Operation: "Increment",
                Value:     2,
            },
        },
    })
```

algolia/search/requests_indexing.go
```diff
@@ -42,3 +42,8 @@ type getObjectsReq struct {
...
+	Operation string      `json:"_operation"`
```
samouss, line 47: We could define a type to restrict the given operations? I don't know how it's done elsewhere in the Go client.

aseure: Mix of both, but we tend not to have dedicated string-aliased types for those, since it would mean having to do a new commit/PR/merge/release every time there's a new value on the engine' side.

PR #613, opened 2020-06-18
feat(rule): implement multi-condition rules

### LINKED ISSUE #614: Multi-Condition Rules feature
This feature brings the support for multi-condition rules, which literally
means you can pass multiple conditions to trigger a single rule. The change
is backward-compatible and would ideally be shipped by mid-July or end-of-July
and it's gonna be deployed by June 19th for us to implement.

1. The feature is about exposing a new `conditions` array of rule conditions
in the rule object. The existing `condition` field remains for
backward-compatibility.
2. Also, please update the CTS like so: https://github.com/algolia/algoliasearch-client-specs-internal/pull/126/files.
3. Mark the `condition` key as deprecated.

aseure:
Close #614

### DIFF
--- a/algolia/search/rule.go
+++ b/algolia/search/rule.go
@@ -11,7 +11,9 @@ import (
 
 // Rule represents an Algolia query rule.
 type Rule struct {
+	// Deprecated: Use `Conditions` instead to specify one or more condition(s)
 	Condition   RuleCondition      `json:"condition"`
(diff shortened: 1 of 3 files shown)


aseure:
@nunomaduro Test updated according to the CTS. Could you please approve? :)

aseure:
@nunomaduro Thanks for reminding me! Just added the deprecation notice.

PR #624, opened 2020-09-21
ExposeIntermediateNetworkErrors

aseure:
**Summary:**

As reported by @sfriquet, the new `opt.ExposeIntermediateNetworkErrors()`
request option was not exposing Algolia engine errors. This was due to
an incorrect assumption that the error returned by the HTTP `.Do()` call
would be non-nil, which isn't in that case (for instance, a 500 does
return a nil error whereas the response itself is a 500 with a response body).

This commit ensures Algolia engine errors are correctly collected when
`opt.ExposeIntermediateNetworkErrors(true)` request option is passed.

**Testing plan:**

A new test has been added: a custom requester, which always return a 500
response without any error, is used to confirm that errors are correctly
reported when `opt.ExposeIntermediateNetworkErrors(true)` is given to
any client function.


aseure:
(copy-pasting here @sfriquet review and answering)

> I'd perhaps argue those are no longer "IntermediateNetworkErrors".

I agree. However, because the type is now exposed, we cannot change it without breaking the compatibility. So I'll have to leave it with this name.

> The default error message can still be misleading, the it can either be a network error or wrong appID/key credentials were used part

True, thanks for raising this point. I've updated it, like we did in the past, to better help users investigate on their own.

sfriquet (approved):

PR #634, opened 2020-11-17
refactor: rename replaceExistingSynonyms into clearExistingSynonyms

### LINKED ISSUE #633: Rename `replaceExistingSynonyms` to `clearExistingSynonyms`
The parameter `replaceExistingSynonyms` in the `saveSynonyms` method is inconsistent with the rest of our API. We should rename it to `clearExistingSynonyms`. Backward-compatibility should be kept in mind

VladislavFitz:
| Q                 | A
| Bug fix?          |  no
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #633 
| Need Doc update   | yes

## Describe your change

[Rename](https://github.com/algolia/algoliasearch-client-specs-internal/issues/149) `replaceExistingSynonyms` into `clearExistingSynonyms` for `saveSynonyms` function.
Keep supporting `replaceExistingSynonyms` as a legacy option.  

Discussed [here](https://github.com/algolia/algoliasearch-client-specs-internal/pull/139)


algolia/internal/gen/options.go
```diff
@@ -155,6 +155,7 @@ var options = []Option{
 	{"query", Other, "", ""},
 	{"autoGenerateObjectIDIfNotExist", Other, false, ""},
 	{"clearExistingRules", Other, false, ""},
+	{"clearExistingSynonyms", Other, false, ""},
 	{"replaceExistingSynonyms", Other, false, ""},
```
rayrutjes, line 159: What happens if both option have contradictory information?

Haroenv: it will take only use replace if clear isn't set https://github.com/algolia/algoliasearch-client-go/pull/634/files#diff-25d7c0cffb7526045febbaea8c194a47dac529e0e14f918ab074774ed8e77554R50-R54

rayrutjes (commented):
It seems to me that the other API clients should be using `replaceExistingSynonyms` and not the other way around.

It looks like the query param on the engine side is `replaceExistingSynonyms`, https://www.algolia.com/doc/api-reference/api-methods/batch-synonyms/#method-param-replaceexistingsynonyms

Going for anything else creates an inconsistency?

Why do we chose to use something different?

VladislavFitz:
I had to put this [link]( https://github.com/algolia/algoliasearch-client-specs-internal/issues/149) in the PR description.

Haroenv (approved):

rayrutjes:
Even reading the different discussions, I don't understand why we would pick a different name than the one of the search engine.
Or at least I would make sure to support the name as present in the API in all API clients.

Haroenv:
we accept both the version from the search engine (replaceExistingSynonyms), as the new clearExistingSynonyms (since it's more correctly named, and more consistent with rules for example). Once we accept both, we update all the documentation to clearExistingSynonyms and the engine's parameter can be changed later to accept both as well.

VladislavFitz:
> Or at least I would make sure to support the name as present in the API in all API clients.
Our goal is to use this format throughout the API clients.

chloelbn (approved):

PR #644, opened 2021-01-29
feat: Dictionaries API methods

VladislavFitz:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no   
| Need Doc update   | yes

## Describe your change

This PR adds the implementation of [Dictionaries API methods](https://www.algolia.com/doc/rest-api/search/#dictionaries-endpoints).


algolia/search/dictionary_entry_compound.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// Compound represents an entry in the Compounds dictionary.
+type Compound struct {
+	objectID      string
+	language      string
+	Word          string
```
chloelbn, line 12: why are the words above with a lower case and the last ones with uppercase (for my personal information)?

algolia/search/responses_dictionaries.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+type SearchDictionariesRes struct {
+	Hits    interface{} `json:"hits"`
+	NbHits  int         `json:"nbHits"`
+	Page    int         `json:"page"`
+	NbPages int         `json:"nbPages"`
+}
+
+func (r SearchDictionariesRes) DictionaryEntires() ([]DictionaryEntry, error) {
```
chloelbn, line 15:
```suggestion
func (r SearchDictionariesRes) DictionaryEntries() ([]DictionaryEntry, error) {
```

chloelbn (changes requested):

algolia/search/dictionary_entry_compound.go
VladislavFitz: The case defines the visibility: upper case is for public, lower case is for the internal usage inside the package. 

I want all the entries to conform to `DictionaryEntry` interface, so I declared `ObjectID()` and `Language()` public functions (getters) for each of them. 

A structure cannot have a public variable and a public function with the same name, that's why they are declared with lower case.

Inspired by Anthony's synonyms implementation.

chloelbn: okay I got it! thanks!

VladislavFitz:
Hey @marchelbling, @samouss , @rayrutjes .
Could please one of you take a look at this PR? (#645 is also pending, btw)
This is my first significant contribution to our Go codebase, so your feedback would be greatly appreciated. 
Thanks in advance!

algolia/search/dictionary_entry_compound.go
```diff
@@ -0,0 +1,51 @@
...
+	Word          string   `json:"word"`
+	Decomposition []string `json:"decomposition"`
+}
+
+func (c Compound) MarshalJSON() ([]byte, error) {
+	return json.Marshal(compound{c.objectID, c.language, c.Word, c.Decomposition})
+}
+
+func (c *Compound) UnmarshalJSON(data []byte) error {
+	if string(data) == jsonNull {
```
netgusto, line 36: If the unmarshaled json byte is actually `null`, shouldn't we have a nil *Compound instead of keeping the current one?

If not, we should at least "nil" the current instance properties as they would remain untouched. I'm aware that it's unlikely someone will unmarshal in an already initialized instance, but I think it highlights the probable need to get a `nil` pointer for unmarshaling a `null`.

algolia/search/dictionary_entry_compound.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot unmarshal Compound dictionary entry: %v", err)
+	}
+
+	c.objectID = compound.ObjectID
+	c.language = compound.Language
+	c.Word = compound.Word
+	c.Decomposition = compound.Decomposition
+	return nil
+}
```
netgusto, line 51: Same remarks on other similar files below:

* dictionary_entry_plural.go
* dictionary_entry_stopword.go

algolia/opt/disable_standard_entries.go
```diff
@@ -0,0 +1,63 @@
...
+package opt
+
+import (
+	"encoding/json"
+	"reflect"
+)
+
+// DisableStandardEntriesOption is a wrapper for an DisableStandardEntries option parameter. It holds
+// the actual value of the option that can be accessed by calling Get.
+type DisableStandardEntriesOption struct {
```
netgusto, line 12: The DisableStandardEntriesOption is contextual to Dictionary, but this is lost in the name/package/code comment. Any way to change this?

algolia/opt/disable_standard_entries.go
```diff
@@ -0,0 +1,63 @@
...
+	value map[string]map[string]bool
```
netgusto, line 13: I noticed there's a type def `DictionaryName` that encodes the enum for this first key of the map. Is it on purpose that this type is not used here?

netgusto (commented):
Solid implementation backed by IT tests, congrats!

I made some comments, mostly questions, that most probably stem from my lack of knowledge of the go search client.

If possible I'd have loved more unit tests on the JSON marshal/unmarshaling logic that seems non trivial (DictionaryEntry, Compound, Plural, ...)

algolia/search/client.go
```diff
@@ -127,3 +127,26 @@ func (c *Client) CustomRequest(
 ) error {
 	return c.transport.Request(&res, method, path, body, k, opts...)
 }
+
+// GetStatus retrieves the task status according to the Algolia engine for the
+// given task.
+func (c *Client) GetStatus(taskID int64) (res TaskStatusRes, err error) {
```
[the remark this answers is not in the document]
VladislavFitz: Good point! These parameters for `GetStatus` and `WaitTask` have been added in [this PR](https://github.com/algolia/algoliasearch-client-go/pull/646). Added them as well for a corresponding client methods.

algolia/search/client_dictionaries.go
```diff
@@ -0,0 +1,62 @@
...
+
+// SearchDictionaryEntries searches for dictionary entries according to the given query string and any rule
+func (c *Client) SearchDictionaryEntries(dictionaryName DictionaryName, query string, opts ...interface{}) (res SearchDictionariesRes, err error) {
+	body := newSearchDictionariesParams(query, opts...)
+	path := c.path("/dictionaries/%s/search", dictionaryName)
+	err = c.transport.Request(&res, http.MethodPost, path, body, call.Read, opts...)
+	return
+}
+
+// GetDictionarySettings Get dictionary settings
```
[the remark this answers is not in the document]
VladislavFitz: This naming has been agreed within common API client specification. I note your remark and we will discuss it.

algolia/search/client_dictionaries.go
```diff
@@ -0,0 +1,62 @@
...
+	return
+}
+
+// ClearDictionaryEntries deletes all the dictionary entries from the give dictonary
+func (c *Client) ClearDictionaryEntries(dictionaryName DictionaryName, opts ...interface{}) (res UpdateTaskRes, err error) {
+	return c.ReplaceDictionaryEntries(dictionaryName, []DictionaryEntry{}, opts)
+}
+
+// SearchDictionaryEntries searches for dictionary entries according to the given query string and any rule
+func (c *Client) SearchDictionaryEntries(dictionaryName DictionaryName, query string, opts ...interface{}) (res SearchDictionariesRes, err error) {
```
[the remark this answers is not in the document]
VladislavFitz: If we have a GET/POST alternative (e.g. [search](https://www.algolia.com/doc/rest-api/search/#search-index-post), we usually expose the POST version as it's more straightforward to encode parameters in the JSON payload than in the URL.

algolia/search/client_dictionaries.go
```diff
@@ -0,0 +1,62 @@
...
+
+// GetDictionarySettings Get dictionary settings
+func (c *Client) GetDictionarySettings(opts ...interface{}) (res DictionarySettings, err error) {
+	path := c.path("/dictionaries/*/settings")
+	err = c.transport.Request(&res, http.MethodGet, path, nil, call.Read, opts...)
+	return
+}
+
+// SetDictionarySettings Set dictionary settings
+func (c *Client) SetDictionarySettings(settings DictionarySettings, opts ...interface{}) (res UpdateTaskRes, err error) {
```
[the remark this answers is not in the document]
VladislavFitz: This was the most difficult part for me.
It actually expects `{"disableStandardEntries": {"stopWords": { /* settings */ } } }`.
I added `DisableStandardEntries` in the [list of options to generate](https://github.com/algolia/algoliasearch-client-go/blob/ab19e602a216d28dc7754488fcfaa61169af1c9f/algolia/internal/gen/options.go#L175) and added a few changes to the generation logic to support this new parameter type (`map[string]map[string]bool`). All other "complex" parameters use the primitive types in their declaration.  I'm not sure if the generator will work with a custom `DictionaryName` type but it makes sense, will try it.

algolia/search/dictionary_entry_compound.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (c Compound) ObjectID() string { return c.objectID }
+func (c Compound) Language() string { return c.language }
+
+// NewCompound is a constructor of Compound
+func NewCompound(objectID, language string, word string, decomposition []string) Compound {
+	return Compound{objectID, language, word, decomposition}
+}
+
+type compound struct {
```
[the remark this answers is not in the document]
VladislavFitz: Here I mimicked the approach used in the [Synonyms interface](https://github.com/algolia/algoliasearch-client-go/blob/c20c1ad93f065ffa3e27db42003baed6f08fc437/algolia/search/synonym.go#L8) implementations (e.g. [one-way](https://github.com/algolia/algoliasearch-client-go/blob/master/algolia/search/synonym_one_way.go), [alt correction](https://github.com/algolia/algoliasearch-client-go/blob/master/algolia/search/synonym_alt_correction_1.go)). Not sure if it's the best way to go, but I prefer to keep the same logic for consistency.

algolia/search/dictionary_entry_compound.go
VladislavFitz: I'm not sure if I understand what do you mean by: 
> have a nil *Compound instead of keeping the current one
Not just return `nil`, but set all the properties to their zero-values/nil? 

Same thing, I borrowed this logic from the [Synonyms unmarshalling code](https://github.com/algolia/algoliasearch-client-go/blob/c20c1ad93f065ffa3e27db42003baed6f08fc437/algolia/search/synonym_alt_correction_1.go#L32) considering it as the best practice.

algolia/opt/disable_standard_entries.go
VladislavFitz: This is a generated code as for all the other properties. I don't think we can group generated properties better than keep them in the `opt` package .

algolia/opt/disable_standard_entries.go
VladislavFitz: As I noticed above, it's a generated property, so I used the embedded types.

algolia/search/dictionary_entry_compound.go
```diff
@@ -0,0 +1,51 @@
...
+
+// Compound represents an entry in the Compounds dictionary.
+type Compound struct {
+	objectID      string
+	language      string
+	Word          string
+	Decomposition []string
+}
+
+func (c Compound) ObjectID() string { return c.objectID }
```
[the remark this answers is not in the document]
VladislavFitz: Fixed ✅

algolia/search/dictionary_entry_compound.go
```diff
@@ -0,0 +1,51 @@
...
+func (c Compound) Language() string { return c.language }
+
+// NewCompound is a constructor of Compound
+func NewCompound(objectID, language string, word string, decomposition []string) Compound {
```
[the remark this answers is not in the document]
VladislavFitz: Fixed ✅

algolia/opt/disable_standard_entries.go
VladislavFitz: **Update**: 
I tried different approaches:
- Move `DictionaryName` to the `opt` package – this introduces a lot of overhead in the generation logic
- Add `DisableStandardEntriesOption` struct manually directly in the `search` package – this introduces a lot of copy-pasted code (raw value wrapper/unwrapper, marshal/unmarshal methods, equals method) that can be generated. And it risks to become even worse if a new per-dictionary parameters will be introduced. 
So, I think that the usage of the raw string as a dictionary name is justified here.

VladislavFitz:
@netgusto Thank you for such a detailed review! I applied the fixes and replied to your questions. Could you reiterate over it?

algolia/opt/disable_standard_entries.go
netgusto: SGTM

algolia/search/dictionary_entry_compound.go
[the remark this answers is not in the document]
netgusto: Understood, thanks!

algolia/search/client_dictionaries.go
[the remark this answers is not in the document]
netgusto: SGTM

algolia/search/client_dictionaries.go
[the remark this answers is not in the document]
netgusto: Have you had success while trying it?

algolia/search/dictionary_entry_compound.go
netgusto: 👌 keeping it in line with the rest of the implem makes sense.

algolia/search/client.go
```diff
@@ -127,3 +127,26 @@ func (c *Client) CustomRequest(
...
+	return
+}
+
+// WaitTask blocks until the task identified by the given taskID is completed on
+// Algolia engine.
+func (c *Client) WaitTask(taskID int64, opts ...interface{}) error {
+	return waitWithRetry(func() (bool, error) {
+		res, err := c.GetStatus(taskID, opts...)
+		if err != nil {
+			return true, err
```
netgusto, line 145: Shouldn't this be better off as `false, err` ? Again, maybe it's similar to other impls, in which case please just disregard.

algolia/search/client.go
```diff
@@ -127,3 +127,26 @@ func (c *Client) CustomRequest(
...
+		}
+		if res.Status == "published" {
```
netgusto, line 147: These 4 lines could be `return res.Status == "published", nil`

netgusto (approved):
LGTM 👍 . Some comments, may be disregarded if irrelevant. Thank you!

algolia/search/client.go
VladislavFitz: Yes, this logic is a copy-paste of `waitWithRetry` for the index. I suppose this was done to not retry inside `waitWithRetry` but to return an error and proceed retrying on the upper level with a different host.

algolia/search/client_dictionaries.go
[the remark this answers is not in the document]
VladislavFitz: Yes, as I mentioned in the [response bellow](https://github.com/algolia/algoliasearch-client-go/pull/644#discussion_r600657185) (first bullet) this requires a lot of changes in the templates generation logic. I don't feel comfortable to introduce such changes and don't think it worth it.

algolia/search/client.go
VladislavFitz: I suppose you meant lines 147-150?
```go
if res.Status == "published" {
  return true, nil
}
return false, nil
```

netgusto: Yes! I'm not sure what happened there :)

VladislavFitz: Done ✅

chloelbn (approved):

PR #649, opened 2021-02-23
feat(Transport): improve error message on unmarshalTo errors

Nagriar:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Improve error message on unmarshalTo errors

When the client is unable to deserialize a response, there's currently no way to know the original message and what it contains. This PR will improve it.

### DIFF
--- a/algolia/transport/transport.go
+++ b/algolia/transport/transport.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"runtime"
@@ -302,10 +303,14 @@ func buildRequest(
 }
 
 func unmarshalTo(r io.ReadCloser, v interface{}) error {
-	err := json.NewDecoder(r).Decode(&v)
+	body, err := ioutil.ReadAll(r)
 	errClose := r.Close()
 	if err != nil {
-		return fmt.Errorf("cannot deserialize response's body: %v", err)
+		return fmt.Errorf("cannot read body: %v", err)
+	}
+	err = json.Unmarshal(body, &v)
+	if err != nil {
+		return fmt.Errorf("cannot deserialize response's body: %v: %s", err, string(body))
 	}
 	if errClose != nil {
 		return fmt.Errorf("cannot close response's body: %v", errClose)
--- a/algolia/transport/transport_test.go
+++ b/algolia/transport/transport_test.go
@@ -1,6 +1,9 @@
 package transport
 
 import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
 	"net/http"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


sfriquet (approved):

algolia/transport/transport.go
```diff
@@ -302,10 +304,14 @@ func buildRequest(
...
 
 func unmarshalTo(r io.ReadCloser, v interface{}) error {
-	err := json.NewDecoder(r).Decode(&v)
+	body, err := ioutil.ReadAll(r)
 	errClose := r.Close()
 	if err != nil {
-		return fmt.Errorf("cannot deserialize response's body: %v", err)
+		return fmt.Errorf("cannot read body: %v", err)
+	}
+	err = json.NewDecoder(bytes.NewReader(body)).Decode(&v)
```
samouss, line 312:
```suggestion
	err = json.Unmarshal(body, &v)
```

We already have the data in `[]byte` we don't have to create a new reader for that. We can call unmarshal.

algolia/transport/transport.go
```diff
@@ -302,10 +304,14 @@ func buildRequest(
...
+	if err != nil {
+		return fmt.Errorf("cannot deserialize response's body: `%s` %v", string(body), err)
```
samouss, line 314:
```suggestion
		return fmt.Errorf("cannot deserialize response's body: %s %v", string(body), err)
```

Not sure the backtick are that useful. If you see a value you could use `%q`.

algolia/transport/transport_test.go
```diff
@@ -79,3 +82,24 @@ func TestShouldExposeIntermediateNetworkErrors(t *testing.T) {
...
+		expectedBody  fakeStruct
+	}{
+		{"<html>Non json answer</html>", fmt.Errorf("cannot deserialize response's body: `<html>Non json answer</html>` invalid character '<' looking for beginning of value"), fakeStruct{}},
+		{`{"attr":"value"}`, nil, fakeStruct{"value"}},
+	} {
+		bodyDeserialized := fakeStruct{}
+		err := unmarshalTo(ioutil.NopCloser(bytes.NewReader([]byte(v.bodyRaw))), &bodyDeserialized)
+		require.Equal(t, v.expectedError, err)
+		fmt.Println(v.expectedBody)
+		fmt.Println(bodyDeserialized)
```
samouss, line 102:
```suggestion
```

algolia/transport/transport.go
Nagriar: The goal is to identify where the body starts and ends. I'll replace it with %q thanks!

samouss: In that case I would probably swap the two `Errorf(body: %v: %s, err, string(body))`. The error is most likely shorter than the entire payload. It would be easier to understand the error and look trough the body after. Otherwise the error message is buried in large blob of text.

samouss (approved):

VladislavFitz (approved):

PR #655, opened 2021-04-01
fix: query parameter support in Search for facet values parameters list

### LINKED ISSUE #654: Query parameter is ignored when searching for facet values
### Description

The search for facet values method supports [passing a variadic list of parameters](https://github.com/algolia/algoliasearch-client-go/blob/c357712b8dc69b685cf0f9a6febeafe96784e3c5/algolia/search/requests_search.go#L85) via generated `QueryParams` structure. `QueryParams` doesn't include `QueryOption`. As a result, `opt.Query` parameters is ignored when searching for facet values. 

### Steps To Reproduce

```go
// The `res` value of this request
res, err := index.SearchForFacetValues("brand", "", opt.Query("abc"), opt.MaxFacetHits(5))

// is identical to the `res` value of this request, even though it doesn't include opt.Query parameter
res, err := index.SearchForFacetValues("brand", "", opt.MaxFacetHits(5))
```

VladislavFitz:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no   
| BC breaks?        | no     
| Related Issue     | Fix #654 
| Need Doc update   | no

## Describe your change

Current implementation of `SearchForFacetValues` method ignores `opt.Query` parameter from the variadic parameters list due to `opt.Query` not being a part of `QueryParams` structure.
This PR adds the `opt.Query` parameter to the `searchForFacetValuesParams` structure and fixes this issue.

### DIFF
--- a/algolia/search/requests_search_test.go
+++ b/algolia/search/requests_search_test.go
@@ -27,3 +27,14 @@ func TestNewSearchParams_ExtraOptionsOverride(t *testing.T) {
 	require.True(t, ok)
 	require.True(t, v)
 }
+
+func TestSearchForFacetValuesParams_IncludeQuery(t *testing.T) {
+	params := newSearchForFacetValuesParams("facet query", opt.Query("search query"), opt.MaxFacetHits(5))
+	data, err := json.Marshal(params)
+	require.NoError(t, err)
+
+	var m map[string]interface{}
+	err = json.Unmarshal(data, &m)
+	require.NoError(t, err)
+	require.NotNil(t, m["query"])
+}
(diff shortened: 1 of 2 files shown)


algolia/search/requests_search_test.go
```diff
@@ -27,3 +27,14 @@ func TestNewSearchParams_ExtraOptionsOverride(t *testing.T) {
...
+
+func TestSearchForFacetValuesParams_IncludeQuery(t *testing.T) {
+	params := newSearchForFacetValuesParams("facet query", opt.Query("search query"), opt.MaxFacetHits(5))
+	data, err := json.Marshal(params)
+	require.NoError(t, err)
+
+	var m map[string]interface{}
+	err = json.Unmarshal(data, &m)
+	require.NoError(t, err)
+	require.NotNil(t, m["query"])
```
netgusto, line 39: Could we also check the value of the query is what we expect?

netgusto (approved):

algolia/search/requests_search_test.go
VladislavFitz: Good point, just added 👍

PR #666, opened 2021-06-22
feat(QueryRes): add support for unknown fields

Jerska:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | no     
| Related Issue     |
| Need Doc update   | no

## Describe your change

Add support for unknown fields in the query response using the special `-` case of `json.Unmarshal`.

In order to debug new fields in the response before support is added in API clients, exposing the unknown fields of the response object would be useful.


Jerska:
Seems harder than I anticipated, as `QueryRes` is unmarshalled in the same place as all other responses: `Transport.Request`.
Closing.

PR #673, opened 2021-06-30
fix: Make Index.Exists() return error if the error is not not-found error

### LINKED ISSUE #672: bug(index): Exists function should return error if the error is not 404
- Algolia Client Version: 3.19.0
- Language Version: 1.16.0

### Description
Currently, `index.Exists` function returns `(false, nil)` even for a not 404 error.
Expected behaivour is below.
```
exists => (true, nil)
not found error =>  (false, nil)
th other errors =>  (false, err)
```

### Steps To Reproduce
```go
validAppID := "XXXXXXXXXX"
ok, err = search.NewClient(validAppID, "invalid").InitIndex(index.GetName()).Exists()
	require.False(t, ok)
         // err shouldn't be nil but 'Algolia API error [403] Invalid Application-ID or API key'
	require.Error(t, err)
```

k-yomo:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | Fix #672  
| Need Doc update   | no

## Describe your change
Fix index.Exists() to return error if the error is not not-found error.

#672

### DIFF
--- a/algolia/search/index.go
+++ b/algolia/search/index.go
@@ -106,10 +106,11 @@ func (i *Index) GetStatus(taskID int64, opts ...interface{}) (res TaskStatusRes,
 // with false.
 func (i *Index) Exists() (bool, error) {
 	_, err := i.GetSettings()
-	_, ok := errs.IsAlgoliaErr(err)
-	if !ok && err != nil {
-		return false, err
+	if err == nil {
+		return true, nil
 	}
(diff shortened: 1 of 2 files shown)


VladislavFitz (approved):

VladislavFitz:
Good catch, thanks @k-yomo !

PR #686, opened 2021-10-20
fix(suggestions) Fix to use struct for fecets in query suggestions config

k-yomo:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | yes     
| Related Issue     | 
| Need Doc update   | no

## Describe your change
Fix to use struct for fecets in query suggestions config for better API interface.

It's not clear for developers what they need to set if we use `map[string]interface{}`.


samouss:
Unfortunately this change is a breaking change... we can't merge it until we perform a next major which is not planned at the moment. We can keep track of that in an issue but that's all we can do at the moment.

PR #688, opened 2022-01-05
fix: reRankingApplyFilter accept same format as facet filters

rayrutjes:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no 
| BC breaks?        | yes     
| Need Doc update   | no

## Describe your change

The `ReRankingApplyFilter` parameter can take the same values as `FacetFilters`.
This PR adjusts the behavior re-using the internal parsing implementation of FacetFilters to support `string`, `[]string` & `[][]string`.

I have removed the comments indicating that the file is generated automatically as this is what was done as well on the face filters side. I guess we only handle simple types with the generation.

I've updated the tests and also removed the go gen comment.

Note that this change is technically a breaking change as the value returned by `ReRankingApplyFilterOption.Get()` is now `[][]string` instead of `string`. However the returned value should always have been `[][]string` as this covers all possibilities.

The current impact being that all users trying to `GetSettings` will observe the error mentioned below.

Currently, when the engine returns something else than a `string`, a `GetSettings` call will fail with the following message:
```
cannot deserialize response's body: cannot unmarshal settings:
json: cannot unmarshal array into Go struct field settings.reRankingApplyFilter of type string:
[...]
```

The payload containing: `{"reRankingApplyFilter":["stock:2"]}` in this instance.


Ant-hem (approved):
Thanks for working on this patch. 

I think that is fine to do a BC because the current implem is already broken. 

Did you try to re-run the options generation to see if it erases `algolia/internal/opt/re_ranking_apply_filter_test.go` ? normally it shouldn't because you removed the header line `// Code generated by go generate. DO NOT EDIT.`, but better be safe that sorry.

shortcuts (approved):
Changes looks fine!

Should we try to release an experimental version before making it public, just in case?

shortcuts:
> it shouldn't because you removed the header line // Code generated by go generate. DO NOT EDIT., but better be safe that sorry.

Agreed, and if we go with a test package it could wait confirmation from @VladislavFitz

Ant-hem:
We can try the generation with:

https://github.com/algolia/algoliasearch-client-go/blob/1ea9c3009b1ee05e075d0adab4dea7a6ab02bae5/Makefile#L3-L4

rayrutjes:
`make generate` did not yield any diff.
I can test this in a limited local setup to check that the GetSettings works as expected on the previously faulty customer.

marchelbling:
Does anyone know why test-1-11 and test-1-12 are still listed and seem to prevent the build to be green entirely?
From what I see we stopped supporting Go 1.11 and 1.12 in https://github.com/algolia/algoliasearch-client-go/pull/679/files.

PR #690, opened 2022-02-17
fix: Suggestion config bool value support

### LINKED ISSUE #671: bug(suggestions): deserialization failed when languages are not set
- Algolia Client Version: master(`v3.19.1-0.20210628231352-f4915f52041a`)
- Language Version: 1.16

### Description
JSON deserialization failed by the wrong value type of `"languages":false` when we get the query suggestions config whose `languages` is not set as following.
Expected behaivour is that the suggestions API returns a json with `"languages": null`. (So this issue is not about client library but API)
https://www.algolia.com/doc/rest-api/query-suggestions/#method-param-languages

```
Error: cannot deserialize response's body: json: cannot unmarshal bool into Go struct field IndexConfiguration.languages of type []string: 
{
  "appID": "XXXXXXXX",
  "sourceIndicesAPIKey": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
  "externalIndicesAPIKey": "",
  "suggestionIndicesAPIKey": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
  "indexName": "test",
  "sourceIndices": [
    {
      "indexName": "test_source",
      "replicas": false,
      "analyticsTags": null,
      "facets": null,
      "minHits": 5,
      "minLetters": 4,
      "generate": null,
      "external": null
    }
  ],
  "languages": false,
  "exclude": null,
  "enablePersonalization": false
}
```

### Steps To Reproduce
1. Create query suggestions index config without setting `languages`.
2. Get the created suggestions index config.
```
package main

import (
	"github.com/algolia/algoliasearch-client-go/v3/algolia/suggestions"
)

func main() {
	appID := "XXXXXXXXXX"
	apiKey := "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"

	searchConfig := search.Configuration{
		AppID:  appID,
		APIKey: apiKey,
	}
	searchClient := search.NewClientWithConfig(searchConfig)
	createIndex(searchClient, "test")
	createIndex(searchClient, "test_source")

	suggestionsConfig := suggestions.Configuration{
		AppID:  appID,
		APIKey: apiKey,
	}
	suggestionsClient := suggestions.NewClientWithConfig(suggestionsConfig)
	err := suggestionsClient.CreateConfig(suggestions.IndexConfiguration{
		IndexName: "test",
			{IndexName: "test_source"},
		},
	})
	if err != nil {
		panic(err)
	}

	_, err = suggestionsClient.GetConfig("test")
	if err != nil {
		panic(err)
	}
}

func createIndex(searchClient *search.Client, indexName string) {
	res, err := searchClient.InitIndex(indexName).SetSettings(search.Settings{})
	if err != nil {
		panic(err)
	}
	if err := res.Wait(); err != nil {
		panic(err)
	}
}
```

VladislavFitz:
| Q                 | A
| Bug fix?          | yes
| New feature?      | yes/no    
| BC breaks?        | no     
| Related Issue     | Fix #671 
| Need Doc update   | no

## Describe your change

This PR introduces `BoolOrStringArray` type to properly represent the query suggestions configuration `languages` value.

Described in #671

### DIFF
--- a/algolia/suggestions/client_config.go
+++ b/algolia/suggestions/client_config.go
@@ -15,7 +15,7 @@ type IndexConfiguration struct {
 	// De-duplicate singular and plural suggestions. Can be either a list languages []string or a boolean.
 	// true value means that all the languages are supported.
 	// false value means that singulars and plurals are not considered the same for matching purposes (foot will not find feet).
-	// []string a list of language ISO codes for which singualr and plural suggestions should be enabled.
+	// []string a list of language ISO codes for which singular and plural suggestions should be enabled.
 	Languages BoolOrStringArray `json:"languages,omitempty"`
 	// List of words and patterns to exclude from the Query Suggestions index.
 	Exclude []string `json:"exclude,omitempty"`
--- a/algolia/suggestions/client_config.go
+++ b/algolia/suggestions/client_config.go
@@ -16,7 +16,7 @@ type IndexConfiguration struct {
 	// true value means that all the languages are supported.
 	// false value means that singulars and plurals are not considered the same for matching purposes (foot will not find feet).
 	// []string a list of language ISO codes for which singular and plural suggestions should be enabled.
-	Languages BoolOrStringArray `json:"languages,omitempty"`
+	Languages BoolOrStringArray `json:"languages"`
 	// List of words and patterns to exclude from the Query Suggestions index.
 	Exclude []string `json:"exclude,omitempty"`
 }
--- a/algolia/suggestions/client_config.go
+++ b/algolia/suggestions/client_config.go
@@ -12,8 +12,11 @@ type IndexConfiguration struct {
 	IndexName string `json:"indexName"`
 	// List of source indices used to generate a Query Suggestions index.
 	SourceIndices []SourceIndex `json:"sourceIndices"`
-	// De-duplicate singular and plural suggestions.
-	Languages []string `json:"languages,omitempty"`
+	// De-duplicate singular and plural suggestions. Can be either a list languages []string or a boolean.
+	// true value means that all the languages are supported.
+	// false value means that singulars and plurals are not considered the same for matching purposes (foot will not find feet).
+	// []string a list of language ISO codes for which singualr and plural suggestions should be enabled.
+	Languages BoolOrStringArray `json:"languages,omitempty"`
 	// List of words and patterns to exclude from the Query Suggestions index.
 	Exclude []string `json:"exclude,omitempty"`
 }
--- a/algolia/suggestions/client_config_test.go
+++ b/algolia/suggestions/client_config_test.go
@@ -1,0 +1,37 @@
+package suggestions
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestConfig_Unmarshal(t *testing.T) {
+	var c IndexConfiguration
+	data := []byte(`{"languages":["ja"]}`)
+	err := json.Unmarshal(data, &c)
+	require.NoError(t, err)
+	require.False(t, c.Languages.IsBool)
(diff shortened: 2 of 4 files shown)


algolia/suggestions/client_config.go
```diff
@@ -12,8 +12,11 @@ type IndexConfiguration struct {
 	IndexName string `json:"indexName"`
 	// List of source indices used to generate a Query Suggestions index.
 	SourceIndices []SourceIndex `json:"sourceIndices"`
-	// De-duplicate singular and plural suggestions.
-	Languages []string `json:"languages,omitempty"`
+	// De-duplicate singular and plural suggestions. Can be either a list languages []string or a boolean.
+	// true value means that all the languages are supported.
+	// false value means that singulars and plurals are not considered the same for matching purposes (foot will not find feet).
+	// []string a list of language ISO codes for which singualr and plural suggestions should be enabled.
```
sfriquet, line 18:
```suggestion
	// []string a list of language ISO codes for which singular and plural suggestions should be enabled.
```

algolia/suggestions/client_config.go
```diff
@@ -12,8 +12,11 @@ type IndexConfiguration struct {
...
+	Languages BoolOrStringArray `json:"languages,omitempty"`
```
sfriquet, line 19: Not sure `omitempty` will work for us here, now that this is no longer nullable: 

```golang
type Foo struct {
	Languages BoolOrStringArray
}

func main() {
	var f Foo
	b, _ := json.Marshal(f)
	fmt.Println(string(b)) // prints {"Languages":null}
}
```

sfriquet (commented):
The PR references an issue with unmarshalling of `languages:false`, can we add a test for this use case ?

algolia/suggestions/client_config.go
VladislavFitz: True, it doesn't work here. But the engine accepts `null` value for the `languages` field. So I just removed the `omitempty` as it is useless.

algolia/suggestions/client_config_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+func TestConfig_Unmarshal(t *testing.T) {
+	var c IndexConfiguration
+	data := []byte(`{"languages":["ja"]}`)
+	err := json.Unmarshal(data, &c)
+	require.NoError(t, err)
+	require.False(t, c.Languages.IsBool)
+	require.Equal(t, []string{"ja"}, c.Languages.StringArray)
+
+	c = IndexConfiguration{}
```
VladislavFitz, line 18: @sfriquet added the test you asked for 👍

algolia/suggestions/client_config_test.go
```diff
@@ -0,0 +1,37 @@
...
+	require.NoError(t, err)
+	require.False(t, c.Languages.IsBool)
+}
+
+func TestConfig_Marshal(t *testing.T) {
+	config := IndexConfiguration{
+		IndexName: "my_index",
+	}
+	json, _ := json.Marshal(config)
+	t.Log(string(json))
```
sfriquet, line 36: what does this test ?

algolia/suggestions/client_config.go
sfriquet: `omitempty` is used when encoding to json, the engine may accept `null` but do you want `null` to be returned ?

VladislavFitz: If I get it right, even if one explicitly sets `languages` to `null`, the engine always return a non-null  (bool or string array) value, so there is no need to support `null`.

algolia/suggestions/client_config_test.go
VladislavFitz: Nice catch, I used it for debugging purposes. Removed.

sfriquet (approved):

PR #691, opened 2022-04-12
feat: update wait task max wait time

sfriquet:
Update wait task logic to not wait for more than 10s between calls, down from 10m, and remove the randomness.
This realigns the go client with the other api clients and improves responsiveness.


Ant-hem (approved):
👍🏻 
note: test suite is flaky on analytics related tests, like: `cts/analytics/aa_testing_test.go`.

BenoitPerrot:
Go go go let's merge that 👍

PR #696, opened 2022-07-11
Allow for SSL to be disabled so we can make non-HTTPS requests

fgribben:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Need Doc update   | yes

## Describe your change

In this PR, we're adding the ability to disable SSL for outgoing requests, to every configuration struct used to instantiate a particular client. This means that, if `DisableSSL` is set to true on the configuration struct, an `http://` request will be made to the target algolia server, rather than an `https://` request.

I want to test parts of my application, of which algolia is a dependency. I found it difficult to create a mock algolia client, so I decided instead to run a mock algolia server implementation, and create a real client in my code that connects to it. The mock implementation I'm using is https://github.com/davesag/mock-algolia. When I run this implementation in docker container, the service runs over `http`, rather than `https`, so, when my test client connects to the docker instance, I get the following intermediate network error:

`http: server gave HTTP response to HTTPS client`

I noticed that, in the mock algolia server repo, it looks as though the Javascript implementation of the algolia client allows you to specify a `http://` scheme, but this doesn't look to be the case for the Go implementation. 

I thought it would be a nice to have `https` as the default scheme, but also allow for `http` if desired. 

Running with this change in place lets me make successful requests to the mock algolia instance in my tests.

### DIFF
--- a/algolia/analytics/client.go
+++ b/algolia/analytics/client.go
@@ -61,6 +61,7 @@ func NewClientWithConfig(config Configuration) *Client {
 			config.Headers,
 			config.ExtraUserAgent,
 			compression.None,
+			config.DisableSSL,
 		),
 	}
 }
--- a/algolia/analytics/configuration.go
+++ b/algolia/analytics/configuration.go
@@ -19,4 +19,5 @@ type Configuration struct {
 	Region         region.Region
 	Headers        map[string]string
 	ExtraUserAgent string
+	DisableSSL     bool
 }
--- a/algolia/insights/client.go
+++ b/algolia/insights/client.go
@@ -50,6 +50,7 @@ func NewClientWithConfig(config Configuration) *Client {
 			config.Headers,
 			config.ExtraUserAgent,
 			compression.None,
+			config.DisableSSL,
 		),
 	}
 }
--- a/algolia/insights/configuration.go
+++ b/algolia/insights/configuration.go
@@ -19,4 +19,5 @@ type Configuration struct {
 	Region         region.Region
 	Headers        map[string]string
 	ExtraUserAgent string
+	DisableSSL     bool
 }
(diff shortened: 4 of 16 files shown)


samouss:
Hello! Thanks for the contribution!

>  I found it difficult to create a mock algolia client, so I decided instead to run a mock algolia server implementation, and create a real client in my code that connects to it. The mock implementation I'm using is https://github.com/davesag/mock-algolia. 

You can define your own interface for the client and hide the Algolia client behind it. You basically define only the methods that you are using. With this approach implementing a mock is trivial, you can have a dumb struct implementing your interface.

> In this PR, we're adding the ability to disable SSL for outgoing requests, to every configuration struct used to instantiate a particular client. This means that, if DisableSSL is set to true on the configuration struct, an http:// request will be made to the target algolia server, rather than an https:// request.

I don't think we have to expose an option for that. The Algolia lets you provide your own [`Requester`](https://github.com/algolia/algoliasearch-client-go/blob/c00d8f8175b3f28135e34306bf931a47f6e419a9/algolia/transport/requester.go#L42-L44) which is basically the object making the request. By default the Algolia client uses an [`http.Client`](https://github.com/algolia/algoliasearch-client-go/blob/c00d8f8175b3f28135e34306bf931a47f6e419a9/algolia/transport/requester.go#L16-L29) behind this interface. You can rely on those two concepts to make a requester using `http` rather than `https` (see below). That's what we are using internally when using the Go client.

```go
// NewHTTPRequester wraps the default HTTP client of the Algolia Go client in a HTTPRequester.
func NewHTTPRequester() *HTTPRequester {
	return &HTTPRequester{Client: transport.DefaultHTTPClient()}
}

// HTTPRequester is a client that enforces the use of HTTP.
type HTTPRequester struct {
	Client *http.Client
}

// Request forwards the input request to the internal client and forces the scheme to be HTTP.
func (r *HTTPRequester) Request(req *http.Request) (*http.Response, error) {
	req.URL.Scheme = "http"
	return r.Client.Do(req)
}

func run() {
	client := search.NewClientWithConfig(search.Configuration{
		AppID:     appID,
		APIKey:    apiKey,
		Requester: searchapi.NewHTTPRequester(),
	})
}
```

PR #698, opened 2022-08-25
chore(circleci): use matrix

samouss:
| Q                 | A
| Bug fix?          | no
| New feature?      | no  
| BC breaks?        | no     
| Related Issue     | no
| Need Doc update   | no

## Describe your change

This PR uses a matrix for the different go version we want to build. It avoids copy/paste the entire job for each version. The idea behind this change is to eventually split the unit from the integration tests. We will be able to run the unit tests on forked PRs. We have change the required checks from `test-1-XX` to `test-1.XX`.

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,7 @@ commands:
             if [ -z "ALGOLIA_ADMIN_KEY_MCM" ]; then
               curl -s https://algoliasearch-client-keygen.herokuapp.com | sh >> $BASH_ENV
             fi
+
   test:
     description: "Run the test for Go << parameters.go_version >>"
     parameters:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -38,91 +39,42 @@ jobs:
...
-            - "~/go/pkg"
-  test-1-17:
+
+  test:
+    parameters:
+        go_version:
+          type: string
     docker:
-      - image: circleci/golang:1.17
+      - image: circleci/golang:<< parameters.go_version >>
```
VladislavFitz, line 55:
```suggestion
      - image: cimg/golang:<< parameters.go_version >>
```
Replace the legacy `circleci/golang` with `cimg/golang`.

samouss: Yep I was going to do it in a follow-up PR, but I can do it now.

samouss:
It looks like it's working! I'm not an admin of the repo but if someone can change the `required` statuses from `test-1-XX` to `test-1.XX` it would be perfect! Next step will be to split the unit/integration to avoid having the check all reds for fork PRs (we can eventually add a build step too).

VladislavFitz (approved):
Great contribution! I updated the required checks. Feel free to merge 👍

PR #699, opened 2022-08-25
fix(indexing): replaceAllObjects blocked if called for non-existent index

### LINKED ISSUE #694: ReplaceAllObjects hangs indefinitely if the index does not exist
```
github.com/algolia/algoliasearch-client-go/v3 v3.25.0
```

I'm pretty sure this has worked at one point.

I'm using it like this:

```go
index := client.InitIndex(indexName)
res, err := index.ReplaceAllObjects(iter, opt.Safe(true))

if err != nil {
	return err
}
return res.Wait()
```

I can see the `tmp` index being created with the correct amount of rows. If I manually create `indexName` before running the above, everything is OK.

My workaround is to use `SaveObjects` for non-existing indices.

VladislavFitz:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #694 
| Need Doc update   | no

## Describe your change

Add a check if the index exists in the `ReplaceAllObjects` function body.
If the index doesn't exist, only the `SaveObjects` operation will be executed which creates the index and saves the provided objects into it.

`ReplaceAllObjects` hangs indefinitely if the index does not exist (#694). This happens due to infinite `CopyIndex` task when called for a non-existent index: the task successfully starts successfully but never accomplishes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
   <p align="center">
     <a href="https://circleci.com/gh/algolia/algoliasearch-client-go"><img src="https://circleci.com/gh/algolia/algoliasearch-client-go.svg?style=shield" alt="CircleCI" /></a>
     <a href="https://github.com/algolia/algoliasearch-client-go/releases"><img src="https://img.shields.io/github/tag/algolia/algoliasearch-client-go.svg" alt="Github Releases"></img></a>
-    <a href="https://godoc.org/github.com/algolia/algoliasearch-client-go"><img src="https://godoc.org/github.com/algolia/algoliasearch-client-go?status.svg" alt="GoDoc"></img></a>
+    <a href="https://pkg.go.dev/github.com/algolia/algoliasearch-client-go/v3"><img src="https://pkg.go.dev/badge/github.com/algolia/algoliasearch-client-go/v3" alt="GoDoc"></img></a>
     <a href="https://goreportcard.com/report/github.com/algolia/algoliasearch-client-go"><img src="https://goreportcard.com/badge/github.com/algolia/algoliasearch-client-go" alt="Go Report Card"></img></a>
     <a href="https://github.com/algolia/algoliasearch-client-go/blob/master/LICENSE"><img src="https://img.shields.io/badge/license-MIT-blue.svg" alt="License"></img></a>
     <img src="https://img.shields.io/badge/Go-%3E=1.11-green.svg" alt="Supported version"></img></a>
--- a/algolia/search/rule_condition.go
+++ b/algolia/search/rule_condition.go
@@ -12,7 +12,7 @@ type RuleCondition struct {
 
 func (c RuleCondition) MarshalJSON() ([]byte, error) {
 	m := make(map[string]interface{})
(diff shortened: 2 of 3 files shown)


algolia/search/index_objects.go
```diff
@@ -344,6 +344,27 @@ func (i *Index) browserForObjects(params searchParams, opts ...interface{}) func
 // prevent issues, one may want to pass the opt.Safe(true) option. Note that
 // passing this option will make the method blocking.
 func (i *Index) ReplaceAllObjects(objects interface{}, opts ...interface{}) (g *wait.Group, err error) {
+
```
samouss, line 347:
```suggestion
```

algolia/search/index_objects.go
```diff
@@ -344,6 +344,27 @@ func (i *Index) browserForObjects(params searchParams, opts ...interface{}) func
...
+	indexExists, err := i.Exists()
+	if err != nil {
+		err = fmt.Errorf("cannot check if the index exists: %v", err)
+		return
+	}
+
+	if !indexExists {
+		resSaveObjects, e := i.SaveObjects(objects, opts...)
+		if e != nil {
+			err = fmt.Errorf("cannot save objects to the index: %v", err)
```
samouss, line 359:
```suggestion
			err = fmt.Errorf("cannot save objects to the index: %v", e)
```

algolia/search/index_objects.go
```diff
@@ -344,6 +344,27 @@ func (i *Index) browserForObjects(params searchParams, opts ...interface{}) func
 // prevent issues, one may want to pass the opt.Safe(true) option. Note that
 // passing this option will make the method blocking.
 func (i *Index) ReplaceAllObjects(objects interface{}, opts ...interface{}) (g *wait.Group, err error) {
+
+	safe := iopt.ExtractSafe(opts...).Get()
+
+	indexExists, err := i.Exists()
```
samouss, line 350:
```suggestion
	exists, err := i.Exists()
```

algolia/search/index_objects.go
```diff
@@ -344,6 +344,27 @@ func (i *Index) browserForObjects(params searchParams, opts ...interface{}) func
...
+	if err != nil {
+		err = fmt.Errorf("cannot check if the index exists: %v", err)
+		return
+	}
+
+	if !indexExists {
```
samouss, line 356:
```suggestion
	if !exists {
```

algolia/search/index_objects.go
```diff
@@ -344,6 +344,27 @@ func (i *Index) browserForObjects(params searchParams, opts ...interface{}) func
...
+		resSaveObjects, e := i.SaveObjects(objects, opts...)
+		if e != nil {
+			err = fmt.Errorf("cannot save objects to the index: %v", err)
+			return
+		}
+		if safe {
+			resSaveObjects.Wait()
```
samouss, line 363: I think we have to check for the returned error of `Wait`.

VladislavFitz: Done ✅

samouss (approved):

PR #702, opened 2022-08-31
fix(rules): Omitempty the facet values rules ordering

clemfromspace:


algolia/search/rendering_content_test.go
```diff
@@ -29,10 +32,15 @@ func TestUnmarshalRenderingContent(t *testing.T) {
...
 	require.Equal(t, r.FacetOrdering.Facets.Order, []string{"brand", "size", "color"})
+
 	require.Equal(t, r.FacetOrdering.Values["brand"].Order, []string{"Apple", "Sony", "Samsung"})
 	require.Equal(t, *r.FacetOrdering.Values["brand"].SortRemainingBy, Alpha)
+
 	require.Equal(t, r.FacetOrdering.Values["color"].Order, []string{"red", "green"})
 	require.Nil(t, r.FacetOrdering.Values["color"].SortRemainingBy)
+
+	require.Nil(t, r.FacetOrdering.Values["size"].Order)
+	require.Equal(t, *r.FacetOrdering.Values["size"].SortRemainingBy, Alpha)
```
samouss, line 43: I'm not sure what this line is testing? With or without the change it doesn't matter.

samouss (commented):
What's the reason behind the change?

clemfromspace:
@samouss Should have been a draft, my bad, was testing things around.

To answer your question: We rely heavily on the Marshal / Unmarshal on the CLI side, and I was trying to copy the rules from one index to another and I got an error like the following:
> Algolia API error [400] Invalid query parameter in consequence: Invalid value for "renderingContent" parameter, expected Error in renderingContent parsing: Unexpected value type: Expecting an array (near 1:647) (in [0]) (near 1:908)
My understanding is that for the rules which do not have an `order` present in the Algolia response, the Go client is setting the `order` to `nil` during the unmarshalling. I will then got a corresponding `"order": null` if during the marshal If I try to import back the rules to Algolia, since the API is expecting either an array for the `order` (even an empty one), or no `order` at all.

The reasoning was: Omit the `order` field if it's`nil`, so it won't be send to Algolia as a `null` JSON value.

Happy to to have your thoughts on this.

algolia/search/rendering_content_test.go
clemfromspace: Indeed, please disregard this change.

samouss:
> My understanding is that for the rules which do not have an `order` present in the Algolia response, the Go client is setting the `order` to `nil` during the unmarshalling. I will then got a corresponding `"order": null` during the marshalling (when I try to import back the rules to Algolia). ...And have the above error, since the API is expecting either an array for the `order` (even an empty one), or no `order` at all (not sure about that though...)

Well, if that's the issue I guess it's easy to test. You can run the CLI with a patched version of the client to see if it works. We could even write an integration tests with the scenario you described. It would ensure that we'll never reproduce the issue.

I don't know what's the process in this case @VladislavFitz: the integration tests are driven by the common test suite. Is it fine to add new integration tests specifically to fix an issue found in the client. I guess it is but prefer to double check!

clemfromspace:
I tested my patch using the Algolia CLI, and it seems to work well with the `omitempty`, although I needed toi add a second one on the `Values` field of the `FacetOrdering` struct because I had a similar error:

> failed to save rules for batch 1398: Algolia API error [400] Invalid query parameter in consequence: Invalid value for "renderingContent" parameter, expected Error in renderingContent parsing: Unexpected value type: Expecting an object.

@VladislavFitz let me know regarding the question of @samouss above, I can add an integration test if necessary.

clemfromspace:
@VladislavFitz @samouss I added one unit test to check the `Marshal` behaviour following my changes. IMHO, it's enough but I can add integrations tests as well if you feel it's necessary.

algolia/search/rendering_content_test.go
```diff
@@ -29,10 +32,92 @@ func TestUnmarshalRenderingContent(t *testing.T) {
...
 	require.Equal(t, *r.FacetOrdering.Values["brand"].SortRemainingBy, Alpha)
+
 	require.Equal(t, r.FacetOrdering.Values["color"].Order, []string{"red", "green"})
 	require.Nil(t, r.FacetOrdering.Values["color"].SortRemainingBy)
+
+	require.Nil(t, r.FacetOrdering.Values["size"].Order)
+}
+
+func TestMarshalRenderingContent(t *testing.T) {
+
```
samouss, line 46:
```suggestion
```

algolia/search/rendering_content_test.go
```diff
@@ -29,10 +32,92 @@ func TestUnmarshalRenderingContent(t *testing.T) {
...
+	"facetOrdering": {
+		"facets": {
+			"order": [
+				"brand",
+				"size",
+				"color"
+			]
+		}
+	}
+}`,
```
samouss, line 72: You can write the JSON with the expected indentation and use `require.JSONEq()` for the assertion.

algolia/search/rendering_content_test.go
```diff
@@ -29,10 +32,92 @@ func TestUnmarshalRenderingContent(t *testing.T) {
...
+			}
+		}
+	}
+}`,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			b, err := json.MarshalIndent(test.input, "", "	")
```
samouss, line 116:
```suggestion
			b, err := json.Marshal(test.input)
```

algolia/search/rendering_content_test.go
```diff
@@ -29,10 +32,92 @@ func TestUnmarshalRenderingContent(t *testing.T) {
...
+			require.NoError(t, err)
+			require.Equal(t, test.expected, string(b))
```
samouss, line 118:
```suggestion
			require.JSONEq(t, test.expected, string(b))
```

VladislavFitz:
@clemfromspace it's ok to add some custom integration tests. The more tests the better 🙂

samouss (approved):

samouss:
Is there anything retaining you from merging the PR? If not please go ahead.

PR #706, opened 2022-10-21
fix(search-account): copy index scope

loicsay:


samouss:
It looks like he `scope` option only exists for the [`search_client.copy_index`](https://github.com/algolia/algoliasearch-client-specs/blob/40796338a0b92feb09afdc404ecd22dcb161b566/api-client-specifications/search_api.md#search_client-interface) function, but not for the [`account_client.copy_index`](https://github.com/algolia/algoliasearch-client-specs/blob/40796338a0b92feb09afdc404ecd22dcb161b566/api-client-specifications/search_api.md#account_client-interface). We may want to stick to the API client specs and avoid implementing a new option not present in other languages. I think @VladislavFitz may have an better view than me on this!

PR #709, opened 2023-01-03
fix(search) hits in search response are now marshalled as 'hits'

aayoubi:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | None
| Need Doc update   | no

## Describe your change

The `hits` property in a search response (`QueryRes`) is now marshalled as `hits` instead of `Hits`.

The `hits` property in a search response (`QueryRes`) was marshalled as `Hits` instead of `hits`, which breaks an integration with InstantSearch in a back-end search implementation. The JS client requires `hits`.


shortcuts (approved):
great catch! thanks :D

millotp (approved):
thanks !

PR #711, opened 2023-01-05
feat(index): add options to index.exists function

### LINKED ISSUE #710: Index.Exists doesn't accept options (and so context)
- Algolia Client Version: 3.26.1

### Description

Currently the function [index.Exists](https://github.com/algolia/algoliasearch-client-go/blob/15d0de75e06af571cff6db73e340f000bf5d278f/algolia/search/index.go#L107) doesn't accept options like other functions.

We use `opts` to pass the current context. With a custom `Requester` (`http.client`) it's allow us to do tracing.
Without opts we can trace correctly this `Exists` function 😞 

Is it possible to add `opts` on `Exists` function, like this ?
```go
func (i *Index) Exists(opts ...interface{}) (bool, error) {
	_, err := i.GetSettings(opts)
	if err == nil {
		return true, nil
	}
	if _, ok := errs.IsAlgoliaErrWithCode(err, http.StatusNotFound); ok {
		return false, nil
	}
	return false, err
}
```

I can create a PR to fix this 👍🏻

k20human:
| Q                 | A
| Bug fix?          | no
| New feature?      | no   (improve an existing function)
| BC breaks?        | no     
| Related Issue     | Fix #710
| Need Doc update   | no

## Describe your change

Currently the function `index.Exists` doesn't accept options like other functions.

Options can be used to propagate the context, for example to do tracing. The goal of this commit is to add `opts` to `index.exists`

Propagate opts to others functions call by `Exists` function

### DIFF
--- a/algolia/search/index_interface.go
+++ b/algolia/search/index_interface.go
@@ -10,7 +10,7 @@ type IndexInterface interface {
 	GetName() string
 	ClearObjects(opts ...interface{}) (res UpdateTaskRes, err error)
 	Delete(opts ...interface{}) (res DeleteTaskRes, err error)
-	Exists() (exists bool, err error)
+	Exists(opts ...interface{}) (exists bool, err error)
 
 	// Indexing
 	GetObject(objectID string, object interface{}, opts ...interface{}) error
--- a/algolia/search/index.go
+++ b/algolia/search/index.go
@@ -104,8 +104,8 @@ func (i *Index) GetStatus(taskID int64, opts ...interface{}) (res TaskStatusRes,
 // Exists returns whether an initialized index exists or not, along with a nil
 // error. When encountering a network error, a non-nil error is returned along
 // with false.
-func (i *Index) Exists() (bool, error) {
-	_, err := i.GetSettings()
+func (i *Index) Exists(opts ...interface{}) (bool, error) {
+	_, err := i.GetSettings(opts...)
 	if err == nil {
 		return true, nil
 	}


algolia/search/index_interface.go
```diff
@@ -10,7 +10,7 @@ type IndexInterface interface {
 	GetName() string
 	ClearObjects(opts ...interface{}) (res UpdateTaskRes, err error)
 	Delete(opts ...interface{}) (res DeleteTaskRes, err error)
-	Exists() (exists bool, err error)
+	Exists(opts ...interface{}) (exists bool, err error)
```
greg0ire, line 13: Isn't this a breaking change?

k20human: No, variadic function can be called with zero or more arguments, so it doesn't change the way we call the function without argument

```
Exists()
Exists(context.Background())
```

samouss: That's correct... Both calls are compatible but the object implementing the interface will complain. It means we cannot merge it because it will break current implementation. The alternative I can think of are:

- introduce a new method like `ExistsWithOpts` and merge the two in the next major
- reimplement the `Exists()` method on your side and expose the options (the function is small)

k20human: Ho yes, totally forgot about this ...

Adding a new `ExistsWithOpts` seems a little hacky i think ... We already reimplement the `Exists` function with options on our side 😉 

So maybe you can add this on your todo list for the next major release ?

samouss: > Adding a new ExistsWithOpts seems a little hacky i think

Yes, it does. 

> We already reimplement the Exists function with options on our side 😉

Perfect.

> So maybe you can add this on your todo list for the next major release ?

Yes. Thanks for your contribution!

samouss:
Will keep this for the next major release.

greg0ire:
@samouss should there be a v4 branch for contributing such breaking changes?

samouss:
@greg0ire We have no plan for such release so far. I've kept the issue with the label `Breaking change`.

Gesisfm:
重複#

PR #712, opened 2023-01-09
fix(search): Add abTestID to search response

cdhawke:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | 
| Need Doc update   | no

## Describe your change

`abTestID` is not currently returned from the client, even though it's a [documented value](https://www.algolia.com/doc/api-reference/api-methods/search/#method-response-abtestid)

This was handled in some of the other clients in the past. e.g. [javascript](https://github.com/algolia/algoliasearch-client-javascript/commit/05059f2)

Returning the value from search responses when the value exists

### DIFF
--- a/algolia/search/responses_search.go
+++ b/algolia/search/responses_search.go
@@ -39,6 +39,8 @@ type QueryRes struct {
 	TimeoutHits                bool                              `json:"timeoutHits"`
 	UserData                   []interface{}                     `json:"userData"`
 	ABTestVariantID            int                               `json:"abTestVariantID"`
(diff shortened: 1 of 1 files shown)


samouss (approved):

cdhawke:
@samouss is it a problem that the integration tests are failing?

samouss:
> @samouss is it a problem that the integration tests are failing?

It's flaky. You can rerun them eventually they should pass.

PR #716, opened 2023-02-08
fix(search) handle automaticFacetFilters unmarshal from string

clemfromspace:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Implement the `UnmarshalJSON` on `AutomaticFacetFilter` to handle the case when `automaticFacetFilters` or `automaticOptionalFacetFilters` are just a `string`.

### DIFF
--- a/algolia/search/rule_consequence_params.go
+++ b/algolia/search/rule_consequence_params.go
@@ -80,6 +80,45 @@ type AutomaticFacetFilter struct {
 	Score       int    `json:"score"`       // Defaults to 1
 }
 
+// AutomaticFacetFilter can be unmarshalled from a string or an object.
+func (a *AutomaticFacetFilter) UnmarshalJSON(data []byte) error {
+	if string(data) == jsonNull {
+		return nil
+	}
+
+	if len(data) == 0 {
+		return nil
+	}
+
+	switch data[0] {
+	case '"':
+		var facet string
+		if err := json.Unmarshal(data, &facet); err != nil {
+			return err
+		}
+		*a = AutomaticFacetFilter{
+			Facet: facet,
+		}
+		return nil
+	case '{':
+		var alias struct {
+			Facet       string `json:"facet"`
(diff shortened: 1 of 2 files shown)


algolia/search/rule_consequence_params.go
```diff
@@ -80,6 +80,45 @@ type AutomaticFacetFilter struct {
...
 
+// AutomaticFacetFilter can be unmarshalled from a string or an object.
+func (a *AutomaticFacetFilter) UnmarshalJSON(data []byte) error {
+	if string(data) == jsonNull {
+		return nil
+	}
+
+	if len(data) == 0 {
+		return nil
+	}
```
samouss, line 91: We can swap. We don't have to cast if the data is empty.

algolia/search/rule_consequence_params.go
```diff
@@ -80,6 +80,45 @@ type AutomaticFacetFilter struct {
...
+			return err
+		}
+		*a = AutomaticFacetFilter{
+			Facet:       alias.Facet,
+			Disjunctive: alias.Disjunctive,
+			Score:       alias.Score,
+		}
+		return nil
+	}
+	return fmt.Errorf("cannot unmarshal automatic facet filter")
```
samouss, line 119: We usually use a `default` case for that.

algolia/search/rule_consequence_params_test.go
```diff
@@ -0,0 +1,33 @@
+package search
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_AutomaticFacetFilter_UnMarshalJSON(t *testing.T) {
```
samouss, line 10:
```suggestion
func Test_AutomaticFacetFilter_UnmarshalJSON(t *testing.T) {
```

shortcuts (approved):
thanks for tackling this!

PR #717, opened 2023-02-16
fix: composableFilterOption deserialization fixed

mehmetaligok:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

`composableFilterOption` deserialization fixed to handle mixed cases. Also, deserialization logic is fixed to take the first level as separate and group. With the changes deserialization handles the following cases;
```
["color:green","color:yellow"] => [["color:green"],["color:yellow"]]

[["color:green","color:yellow"], "color:blue"] => [["color:green","color:yellow"], ["color:blue"]]
```

Other client library PRs that fix related issues are based.  Ex: https://github.com/algolia/algoliasearch-client-java-2/pull/771

**PS:** When fixing this issue, we noticed that legacy one string parsing is also not handled properly in this client. It will be addressed with another PR.

### DIFF
--- a/algolia/opt/composable_filter.go
+++ b/algolia/opt/composable_filter.go
@@ -2,6 +2,7 @@ package opt
 
 import (
 	"encoding/json"
+	"fmt"
 	"reflect"
 	"strings"
 
@@ -74,10 +75,11 @@ func (o *composableFilterOption) UnmarshalJSON(data []byte) error {
 	}
 
 	var (
-		ok     = false
-		filter string
-		ors    []string
-		ands   [][]string
+		ok      = false
+		filter  string
+		ors     []string
+		ands    [][]string
+		options []interface{}
 	)
 
 	if json.Unmarshal(data, &filter) == nil {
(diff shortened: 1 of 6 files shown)


algolia/opt/composable_filter.go
```diff
@@ -86,13 +88,25 @@ func (o *composableFilterOption) UnmarshalJSON(data []byte) error {
...
 
-	if !ok && json.Unmarshal(data, &ands) == nil {
-		ok = true
+		for _, option := range options {
+			switch v := option.(type) {
+			case []interface{}:
+				ors = []string{}
+
+				for _, val := range v {
+					ors = append(ors, fmt.Sprint(val))
```
sbellone, line 100: Why is `fmt.Sprint` needed?

sbellone (approved):
Looks great, thanks for tackling it!

algolia/opt/composable_filter.go
mehmetaligok: Since we need to convert `[]interface{}` to `[]string`, I needed to convert `val`(interface{}) to string. Type conversion could also be done, but this approach seemed simpler to me.

sbellone: Oh, of course :ok_hand:

PR #718, opened 2023-02-23
fix: handle composableFilterOption deserialization for legacy usage

mehmetaligok:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

One-string legacy filter deserialization behavior is fixed to handle parenthesis usage with multiple groups.
Added more tests to cover possible usages.

**PS:** This behavior is already fixed in other client libraries. This implementation is based on those PRs which are fixing the same issue.

https://github.com/algolia/algoliasearch-client-java-2/pull/771
https://github.com/algolia/algoliasearch-client-csharp/pull/803

When legacy one-string filters are used deserialization was not working as expected. 
Normally, it should work as below;
```
"color:green,color:yellow" => [["color:green"],["color:yellow"]]

"(color:green,color:yellow),color:blue" => [["color:green","color:yellow"], ["color:blue"]]

"(color:green,color:yellow)" => [["color:green","color:yellow"]]
```

### DIFF
--- a/algolia/opt/composable_filter.go
+++ b/algolia/opt/composable_filter.go
@@ -139,7 +139,7 @@ func (o *composableFilterOption) UnmarshalJSON(data []byte) error {
 		for _, filter := range ors {
 			filter = strings.Trim(filter, " ")
 			if len(filter) > 0 {
-				cleanORs = append(cleanORs, strings.Split(filter, ",")...)
+				cleanORs = append(cleanORs, filter)
 			}
 		}
 		if len(cleanORs) > 0 {
--- a/algolia/opt/composable_filter.go
+++ b/algolia/opt/composable_filter.go
@@ -82,6 +82,11 @@ func (o *composableFilterOption) UnmarshalJSON(data []byte) error {
 		options []interface{}
 	)
 
+	// Handles legacy one-string format as filters.
+	// Adds outer groups as `ands`, adds inner groups as `ors` if there are any.
+	//"A:1,B:2"       => [["A:1"],["B:2"]]
+	//"(A:1,B:2),C:3" => [["A:1","B:2"],["C:3"]]
+	//"(A:1,B:2)"     => [["A:1","B:2"]]
 	if json.Unmarshal(data, &filter) == nil {
 		ok = true
 		replacer := strings.NewReplacer("(", " ", ")", " ")
--- a/algolia/opt/composable_filter.go
+++ b/algolia/opt/composable_filter.go
@@ -84,7 +84,27 @@ func (o *composableFilterOption) UnmarshalJSON(data []byte) error {
 
 	if json.Unmarshal(data, &filter) == nil {
 		ok = true
-		ors = strings.Split(filter, ",")
+		replacer := strings.NewReplacer("(", " ", ")", " ")
+
+		var start, count int
+		for i, c := range filter {
+			switch c {
+			case '(':
+				count++
+			case ')':
+				count--
+			case ',':
+				if count == 0 {
+					// remove parentheses and split filters by comma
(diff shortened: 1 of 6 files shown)


algolia/opt/composable_filter_test.go
```diff
@@ -84,6 +84,20 @@ func TestComposableFilterOption_UnmarshalJSON(t *testing.T) {
...
+			composableFilterOption{[][]string{
+				{`color:green,color:yellow`},
+				{`color:blue`},
+			}},
+		},
+		{
+			`"(color:green,color:yellow),color:black"`,
+			composableFilterOption{[][]string{
+				{`color:green`, `color:yellow`},
+				{`color:blue`},
```
sbellone, line 98:
```suggestion
				{`color:black`},
```
Should we test the payload also?

algolia/opt/composable_filter.go
```diff
@@ -84,7 +84,27 @@ func (o *composableFilterOption) UnmarshalJSON(data []byte) error {
 
 	if json.Unmarshal(data, &filter) == nil {
 		ok = true
-		ors = strings.Split(filter, ",")
+		replacer := strings.NewReplacer("(", " ", ")", " ")
```
sbellone, line 87: What about the `[`, I see that it's supported in the tests, are they handled elsewhere?

sbellone (commented):
Looks good! :ok_hand: 
Maybe we can add a comment to understand easily what it does without having to look at the tests? Like what you've put in this PR desc

algolia/opt/composable_filter_test.go
mehmetaligok: Let me check this out. I saw that other related tests(facet, tags etc...) are checking the payload as well but this one seems not. 
I am not sure if there was a reason but if it will be possible, I will try to add that to make tests stronger.

algolia/opt/composable_filter.go
mehmetaligok: Yes, they are handled as slices because `Unmarshal` will not unmarshal that as a string and they will be handled in the below.

algolia/opt/composable_filter_test.go
mehmetaligok: It looks like the test was only comparing `len` of the slices. Removing that did the trick. 

https://github.com/algolia/algoliasearch-client-go/pull/718/commits/734da0f4f7c2e1b7c3fed70a318660b4eebc770b

algolia/opt/composable_filter.go
sbellone: Ahh of course :)

algolia/opt/composable_filter_test.go
sbellone: But `black` and `blue` are not of the same length so it should have failed :thinking:

mehmetaligok: It wasn't checking the actual value length. It was checking slice length only. For example, the expected slice has a length of 2 and the actual slice has a length of 2. Please see the picture below, `fGot` and `fExpected` are `[][]string`. 

image


But with the change that I do now, it should check the whole slice with the elements. So it only passed after I changed `black` to `blue`.

sbellone: Thanks for the explanation, I should have re-read everything!

sbellone (approved):
:rocket:

algolia/opt/composable_filter_test.go
mehmetaligok: no worries at all, thanks for following up 🙌

PR #720, opened 2023-03-28
fix: set default value to string slices to prevent `null` unmarshalling

mehmetaligok:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Added default value to string slice initialization to prevent `null` unmarshalling. 

When initializing string slices with empty arguments, the value was set to nil, and when unmarshalling it option value was sent as `null` instead `[]`.

Example;  

```
res, err := index.SetSettings(search.Settings{
		Replicas: opt.Replicas(),
	})
```

sent as 
`{"replicas": null}` instead of `{"replicas": []}`. 

This PR is fixing the this with updating `string_slice.go.tmpl`.


shortcuts (approved):
looks great, nice find!

don't we have tests for this somehow? feel like a unit would assert it properly

PR #725, opened 2023-04-10
feat(abtests): add index_prefix and index_suffix options for ab_testing

cdhawke:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Need Doc update   | yes

## Describe your change

The analytics API recently added support for `indexPrefix` and `indexSuffix` query parameters on the LIST (GET) endpoint. This allows filtering of variants by their associated index name prior to retrieval. This change adds the `indexPrefix` and `indexSuffix` option extraction to the ab testing client.

Related Update: https://github.com/algolia/go/pull/10733

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,11 +101,11 @@ workflows:
           matrix:
               parameters:
                 go_version:
-                    - "1.13"
-                    - "1.14"
-                    - "1.15"
                     - "1.16"
                     - "1.17"
+                    - "1.18"
+                    - "1.19"
+                    - "1.20"
 
       - integration-tests:
           filters:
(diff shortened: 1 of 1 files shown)


algolia/internal/gen/options.go
```diff
@@ -1,3 +1,4 @@
+//go:build ignore
```
JonathanMontane, line 1: While I agree that this is how new tags should be written, do we need this?
I believe the previous syntax (on the line below) is still valid go.

JonathanMontane (approved):

JonathanMontane (approved):

algolia/internal/gen/options.go
JonathanMontane: No worries, just wanted to make sure I wasn't missing something overly subtle.

JonathanMontane (approved):

PR #726, opened 2023-04-11
feat: allow to set a custom delay for the wait functions

Nagriar:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Allow to set a custom delay for the wait functions using the options

In case of low latency, the default 1s delay can be too high.


algolia/internal/opt/wait_configuration.go
```diff
@@ -0,0 +1,16 @@
+package opt
+
+import (
+	"github.com/algolia/algoliasearch-client-go/v3/algolia/opt"
+)
+
+// ExtractScopes returns the first found ScopesOption from the
```
Ant-hem, line 7:
```suggestion
// ExtractWaitConfiguration returns the first found wait configuration from....
```

algolia/opt/wait_configuration.go
```diff
@@ -0,0 +1,33 @@
...
+
+// WaitConfigurationOption stores the logic to change the delay between waits
+type WaitConfigurationOption struct {
+	DelayGrowth func(*time.Duration) time.Duration
+}
+
+// DefaultWaitConfiguration returns the new delay to apply between waits
+func DefaultWaitConfiguration() *WaitConfigurationOption {
+	return &WaitConfigurationOption{DelayGrowth: func(d *time.Duration) time.Duration {
+		const maxDurationBetweenWaits = 10 * time.Second
```
Ant-hem, line 15: could be added at the top level of the file to be consistent with the code base.

Ant-hem (commented):
thanks - few styling comment

PR #735, opened 2023-05-15
fix(recommend): Don't return error when unknown scope for taskID

dkurzaj:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Don't return an error when an unknown scope has been detected for a given taskID. Instead, simply return an empty string scope.

In some cases we might be confronted to taskIDs with unknown scopes, and thus we should not fail and return an error.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,11 +1,5 @@
 # ChangeLog
 
-## [3.29.3](https://github.com/algolia/algoliasearch-client-go/compare/v3.29.2...v3.29.3) (2023-05-15)
-
-## Fix
-
-- fix(recommend): Don't return error when unknown scope for taskID (#735) ([c1be0f7](https://github.com/algolia/algoliasearch-client-go/commit/c1be0f7))
-
 ## [3.29.2](https://github.com/algolia/algoliasearch-client-go/compare/v3.29.1...v3.29.2) (2023-05-11)
 
 ## Fix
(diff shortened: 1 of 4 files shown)


jojva (commented):
I don't know if this is the right way to do this... What does "empty scope" mean? Should we give a default scope instead (thinking about either `index` or `metis`)? Since I don't really know how Metis taskIDs work, I'm not confident in how they should be handled.

Have you found any documentation regarding Metis taskIDs?

dkurzaj:
> I don't know if this is the right way to do this... What does "empty scope" mean? Should we give a default scope instead (thinking about either `index` or `metis`)? Since I don't really know how Metis taskIDs work, I'm not confident in how they should be handled.
> 
> Have you found any documentation regarding Metis taskIDs?

I don't really know, but at least it's keeping the same behaviour as before https://github.com/algolia/algoliasearch-client-go/pull/730 for all the non-Recommend taskIDs 🤔

dkurzaj:
> I don't know if this is the right way to do this... What does "empty scope" mean? Should we give a default scope instead (thinking about either `index` or `metis`)? Since I don't really know how Metis taskIDs work, I'm not confident in how they should be handled.
> 
> Have you found any documentation regarding Metis taskIDs?

I've modified the PR in order to not modify the `getScopeFromTaskID` method and keep the scope retrieval mechanism. But simply ignored the error if it's not a recommend scope in order to keep the same behaviour as before for the other taskIDs.

xiangb (approved):

dkurzaj:
Won't do.

PR #736, opened 2023-05-16
fix(recommend): implement dedicated waitRecommendTask function

jojva:
| Q                 | A
| Bug fix?          | yes
| New feature?      | yes    
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Add a `WaitRecommendTask` function.

Fix `waitTask` implementation for Recommend: use a dedicated function rather than relying on the `taskID` value, which is computed differently in Metis.

### DIFF
--- a/algolia/search/client_interface.go
+++ b/algolia/search/client_interface.go
@@ -9,7 +9,9 @@ import (
 type ClientInterface interface {
 	// Misc
 	WaitTask(taskID int64, opts ...interface{}) error
+	WaitRecommendTask(taskID int64, opts ...interface{}) error
 	GetStatus(taskID int64, opts ...interface{}) (res TaskStatusRes, err error)
+	GetRecommendStatus(taskID int64, opts ...interface{}) (res TaskStatusRes, err error)
 	InitIndex(indexName string) *Index
 	ListIndices(opts ...interface{}) (res ListIndicesRes, err error)
 	GetLogs(opts ...interface{}) (res GetLogsRes, err error)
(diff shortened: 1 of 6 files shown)


dkurzaj (approved):
LGTM

dkurzaj (approved):

algolia/search/client_interface.go
```diff
@@ -9,7 +9,9 @@ import (
 type ClientInterface interface {
 	// Misc
 	WaitTask(taskID int64, opts ...interface{}) error
+	WaitRecommendTask(taskID int64, opts ...interface{}) error
```
samouss, line 12: This is a breaking change. It should have been done in a major. Client implementing the interface will now receive a compilation error because their implementation is lacking `WaitRecommendTask` & `GetRecommendStatus`.

algolia/search/index_interface.go
```diff
@@ -5,7 +5,9 @@ import "github.com/algolia/algoliasearch-client-go/v3/algolia/wait"
 type IndexInterface interface {
 	// Misc
 	WaitTask(taskID int64, opts ...interface{}) error
+	WaitRecommendTask(taskID int64, opts ...interface{}) error
```
samouss, line 8: This is a breaking change similar to the client interface.

algolia/search/client_interface.go
samouss: Worth asking the DX team imo. Not sure who is responsible for this package anymore.

jojva: We don't really know either. I'll ask in #help-apiclients.

PR #737, opened 2023-06-06
feat(renderingContent): parse the redirect attribute

Nagriar:
| Q                 | A
| Bug fix?          | yes
| New feature?      | yes
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Parse the `redirect` attribute and prevent to serialize `FacetOrdering` when nil

When unmarshalling a rendering content with only `redirect`, the attribute was ignored. Then marshalling the struct resulted in an invalid renderint content because `FacetOrdering` set to nil is invalid


Ant-hem (approved):
thanks!

PR #738, opened 2023-07-04
fix(transport): allow to override existing header with different case

Nagriar:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Allow to override existing header with different case instead of appending values

The current behaviour append header values but the search engine doesn't support multivalued headers. It leads to undefined behaviour based on the header ordering.

### DIFF
--- a/algolia/transport/transport.go
+++ b/algolia/transport/transport.go
@@ -26,7 +26,7 @@ const version = "3.30.0"
 type Transport struct {
 	requester     Requester
 	retryStrategy *RetryStrategy
-	headers       map[string]string
+	headers       http.Header
 	compression   compression.Compression
 }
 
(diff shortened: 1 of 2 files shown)


Ant-hem (approved):

algolia/transport/transport.go
```diff
@@ -26,7 +26,7 @@ const version = "3.30.0"
 type Transport struct {
 	requester     Requester
 	retryStrategy *RetryStrategy
-	headers       map[string]string
+	headers       http.Header
```
samouss, line 29: It's a breaking change no? We could add another field taking precedence if provided.

Ant-hem: fwiw, in the past we used to accept BC on bug fixes.

Nagriar: > It's a breaking change no? We could add another field taking precedence if provided.

So before, if we were playing with difference case, we were appending values together while now it overrides following this precedence:
- Request header
- Extra client header
- Default client header

samouss (approved):

PR #740, opened 2023-08-24
fix(rules): missing redirect field in RuleConsequence

dhowden:
Previously the SDK would silently drop fields specific to redirect-based rules.  This change adds the missing fields so that it is possible to fetch and re-submit rules without losing the redirect information.

Ideally the SDK would support this natively for all types, but that is a bigger job!


trezz (approved):

Ant-hem:
Could you please update `algolia/search/rule_test.go` and enrich the PR's description?

Ant-hem (approved):

PR #741, opened 2023-08-25
fix(circleci): integration-tests must not run in parallel

dhowden:
The integration tests for each Go version are using the same underlying resources and so are triggering races which causes the tests to fail.


dhowden:
Closing due to age

PR #753, opened 2024-02-28
fix(transport): clone headers when creating a request

helenamariano:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no 
| BC breaks?        | no     
| Related Issue     | 
| Need Doc update   | no

## Describe your change
This change clones the http headers when making a new  http request. Previously the same instance of the header map would be used for all requests.

This fixes a panic due to concurrent map write that can occur if you have a custom `transport.Requester` that adds additional headers to the request.

### DIFF
--- a/algolia/transport/transport.go
+++ b/algolia/transport/transport.go
@@ -87,7 +87,7 @@ func (t *Transport) Request(
 ) error {
 	var (
 		ctx       = iopt.ExtractContext(opts...)
-		headers   = t.headers
+		headers   = t.headers.Clone()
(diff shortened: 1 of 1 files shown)


dhowden:
LGTM (we paired on this fix).

Those tests appear unrelated (they seem to be flakey).  I've tried to correct this before (https://github.com/algolia/algoliasearch-client-go/pull/741) but didn't get very far.  We should look into this on a separate issue.

dhowden (approved):

PR #759, opened 2024-04-23
Do not return error when trying to unmarshal an empty response body

### LINKED ISSUE #758: Unable to Create/Update suggestion index configuration
- Algolia Client Version: 3.31.1
- Language Version: 1.22

### Description
When trying to create or update a suggestion index configuration, we get an error saying that the JSON response body could not be deserialized.
Our api key has the correct ACLs according to the documentation. We even tried with granting it all rights and still get the same error.

image

### Steps To Reproduce
```go
suggestionsClient := suggestions.NewClient(appID, apiKey)

minHits := 2
minLetters := 2

config := suggestions.IndexConfiguration{
	IndexName: "test_query_suggestions",
		{
			IndexName:  "test",
			MinHits:    &minHits,
			MinLetters: &minLetters,
		},
	},
	Languages: suggestions.BoolOrStringArray{
		IsBool:      false,
		StringArray: []string{language},
	},
}

err := suggestionsClient.CreateConfig(config, opt.ExposeIntermediateNetworkErrors(true))
```

We hit this breakpoint in transport.go in the Algolia code:
image
The response body looks like it is empty and unmarshalling the empty body fails?

image

The calls `suggestionsClient.ListConfigs()` and `suggestionsClient.GetConfig()` do work, so we know the setup of the client works.

**Update:**
The real error is not the parsing of the response body, but it goes wrong earlier.
We get a 307 response from the POST endpoint: 
image

Things seem to go wrong in the wrapJSONEncoder function.

andersaction:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #758  
| Need Doc update   | no

## Describe your change
When doing calls like `CreateConfig` and `UpdateConfig` it seems that the response body is empty and the function `unmarshalTo` returns an error. I think an empty body for an api call that does not return anything should not cause an error.

https://www.algolia.com/doc/rest-api/query-suggestions/#create-a-configuration
https://www.algolia.com/doc/rest-api/query-suggestions/#update-a-configuration

Please let me know if this solution makes sense and if it could have any other unwanted consequences.

As described in the issue, we are not able to make these API calls successfully because of an error returned from the `unmarshalTo` function which seems to be caused by trying to unmarshal an empty body.

### DIFF
--- a/algolia/transport/transport.go
+++ b/algolia/transport/transport.go
@@ -308,6 +308,9 @@ func unmarshalTo(r io.ReadCloser, v interface{}) error {
 	if err != nil {
 		return fmt.Errorf("cannot read body: %v", err)
 	}
+	if len(body) == 0 {
+		return nil
+	}
 	err = json.Unmarshal(body, &v)
 	if err != nil {
 		return fmt.Errorf("cannot deserialize response's body: %v: %s", err, string(body))
--- a/algolia/transport/transport_test.go
+++ b/algolia/transport/transport_test.go
@@ -96,6 +96,7 @@ func TestUnmarshallTo(t *testing.T) {
 	}{
 		{"<html>Non json answer</html>", fmt.Errorf("cannot deserialize response's body: invalid character '<' looking for beginning of value: <html>Non json answer</html>"), fakeStruct{}},
 		{`{"attr":"value"}`, nil, fakeStruct{"value"}},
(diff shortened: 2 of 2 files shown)


andersaction:
This does not seem like the real case of the error. Investigating some more and might open a different PR for it.

PR #760, opened 2024-04-24
Make requests with body in the Suggestions API work

### LINKED ISSUE #758: Unable to Create/Update suggestion index configuration

andersaction:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no 
| BC breaks?        | no     
| Related Issue     | Fix #758 
| Need Doc update   | no

## Describe your change
This change in the `wrapJSONEncoder` function makes the POST and PUT calls in the Suggestions API work without failure.

Interestingly other calls with a body from the search client like MultipleQueries work fine using the old `wrapJSONEncoder`, which leads me to think that the problem might be in the struct IndexConfiguration, which is the body of our failing calls.
image

Either way this change seems to work for both calls from suggestions client and search client.

As described in the issue, CreateConfiguration and UpdateConfiguration does not work for us: 
https://www.algolia.com/doc/rest-api/query-suggestions/#create-a-configuration
https://www.algolia.com/doc/rest-api/query-suggestions/#update-a-configuration

### DIFF
--- a/algolia/transport/transport.go
+++ b/algolia/transport/transport.go
@@ -1,6 +1,7 @@
 package transport
 
 import (
+	"bytes"
 	"compress/gzip"
 	"context"
 	"encoding/json"
@@ -204,22 +205,16 @@ func mergeHeaders(defaultHeaders http.Header, extraHeaders map[string]string) ht
 	return headers
 }
 
-func wrapJSONEncoder(in interface{}) io.ReadCloser {
-	pr, pw := io.Pipe()
-	go func() {
-		errEncode := json.NewEncoder(pw).Encode(in)
-		errClose := pw.Close()
-		if errEncode != nil {
-			debug.Printf("cannot JSON encode request body: %v", errEncode)
-		}
-		if errClose != nil {
-			debug.Printf("cannot close JSON encoder writer for request body: %v", errClose)
-		}
-	}()
-	return pr
+func wrapJSONEncoder(in interface{}) io.Reader {
+	b := new(bytes.Buffer)
+	errEncode := json.NewEncoder(b).Encode(in)
+	if errEncode != nil {
+		debug.Printf("cannot JSON encode request body: %v", errEncode)
+	}
+	return b
 }
 
-func wrapGZIPEncoder(in io.ReadCloser) io.ReadCloser {
+func wrapGZIPEncoder(in io.Reader) io.Reader {
 	pr, pw := io.Pipe()
 	go func() {
 		gw := gzip.NewWriter(pw)
@@ -244,7 +239,7 @@ func buildRequestWithoutBody(method, url string) (*http.Request, error) {
 }
 
 func buildRequestWithBody(method, url string, body interface{}, c compression.Compression) (*http.Request, error) {
-	var r io.ReadCloser
+	var r io.Reader
 	jsonEncoder := wrapJSONEncoder(body)
 	switch c {
 	case compression.GZIP:


nicknijenhuis (approved):
LGTM!

shortcuts (commented):
The solution seems to work however I believe the pipe implementation allows more robust memory allocation

It seems like the underlying issue you highlight here https://github.com/algolia/algoliasearch-client-go/issues/758 is just that the response is empty, no? In that case we can maybe early return if `in` is nil?

cc @Nagriar if you remember the reason why we picked this implem

shortcuts:
(Thanks for taking the time to investigate btw!!)

hernanif1:
Hi @shortcuts, I've been working with @andersaction to implement this PR

before:
```go
	pr, pw := io.Pipe()
	go func() {
		errEncode := json.NewEncoder(pw).Encode(in) // in is not empty but the pr return body 0
	...
```

even when `in` has content, this code is not parsing the body correctly resulting the `pr` with body 0 

I don't see how an earlier return could help in that case.
But I imagine that this code works for you. 

I wonder if it is our Go Lang version 1.22 that might handle the io.Pipe differently than 1.13.

shortcuts:
> Hi @shortcuts, I've been working with @andersaction to implement this PR
> 
> before:
> 
> ```go
> 	pr, pw := io.Pipe()
> 	go func() {
> 		errEncode := json.NewEncoder(pw).Encode(in) // in is not empty but the pr returns body 0
> 	...
> ```
> 
> even when `in` has content, this code is not parsing the body correctly resulting the `pr` with body 0
> 
> I don't see how an earlier return could help in that case. (or I didn't get it, also possible)
> 
> But I imagine that this code (current code without modification) works for you. I wonder if it is our Go Lang version 1.22 that might handle the io.Pipe differently than 1.13.

Oh ok I thought it was empty, I did not tried to reproduce the error yet but for the next major version of the client (available on https://github.com/algolia/algoliasearch-client-go/tree/next) we've went with a similar approach so I'd like to know if there was motivations to use Pipe in the past

hernanif1:
@shortcuts I'll install the "next" version tomorrow and I'll let you know, I'm not sure if it will work smoothly, so let's see 😅

shortcuts:
> @shortcuts I'll install the "next" version tomorrow and I'll let you know, I'm not sure if it will work smoothly, so let's see 😅

yup there is a pretty huge number of breaking changes.. you can find the docs for the beta here https://api-clients-automation.netlify.app/ if you want, but I'll follow up with this PR asap too :)

andersaction:
> The solution seems to work however I believe the pipe implementation allows more robust memory allocation
> 
> It seems like the underlying issue you highlight here #758 is just that the response is empty, no? In that case we can maybe early return if `in` is nil?
> 
> cc @Nagriar if you remember the reason why we picked this implem

At first I thought the problem was with the empty response, but upon further investigation it seems like it goes wrong already when serializing the request body to JSON. 

Since other POST calls from other clients work, I wonder if there could be something wrong with the `IndexConfiguration` struct itself (or `SourceIndex` or `BoolOrStringArray`) in combination with the implementation of `wrapJSONEncoder`.

hernanif1:
Hi @shortcuts, just to let you know that we are handling the config manually, so we are covered for now, but updating them manually is a nightmare 😅. So no big rush, but please notify me when you have a forecast of a possible date to have a look, if you need additional information or some concerns about this so we can update it.  Glad to help. 😀

shortcuts:
(Sorry for the delay I had a lot of stuff to tackle)

Hey! So I've tried some stuff with the client in order to reproduce the error and couldn't make it fail, but then I saw you mention 307 redirect not being caught and then I managed to make the request fail by misconfiguring (or leaving the default, `US`, value) my region.

My app region is in `EU` (as is yours), when I try to call the client with no region set (defaults to `US`), I get `panic: all hosts have been contacted unsuccessfully, it can either be a server or a network error or wrong appID/key credentials were used. You can use opt.ExposeIntermediateNetworkErrors(true) to investigate.`, but when I properly set it to `EU` I can call the API to create/update config without issue

Would you letting me know if defining the region works for you? or if not, what is the error you exactly get from the API call?

Below is attached the program I've ran, `testQs` should match your implem I believe

```go
package main

import (
	"fmt"

	"github.com/algolia/algoliasearch-client-go/v3/algolia/opt"
	"github.com/algolia/algoliasearch-client-go/v3/algolia/region"
	"github.com/algolia/algoliasearch-client-go/v3/algolia/suggestions"
)

func testQs() int {
	client := suggestions.NewClientWithConfig(suggestions.Configuration{
		AppID:  "QPBQ67WNIG",
		APIKey: "",
		Region: region.EU,
	})

	minHits := 2
	minLetters := 2

	config := suggestions.IndexConfiguration{
		IndexName: "testbar",
			{
				IndexName:  "api-clients-automation",
				MinHits:    &minHits,
				MinLetters: &minLetters,
			},
		},
		Languages: suggestions.BoolOrStringArray{
			IsBool:      false,
			StringArray: []string{"en"},
		},
	}

	err := client.CreateConfig(config, opt.ExposeIntermediateNetworkErrors(true))
	if err != nil {
		panic(err)
	}

	return 0
}
```

andersaction:
@shortcuts thanks for diving into it. Setting the region to EU seems to fix the issue indeed. Sorry for missing that. Closing this PR and the issue.

A nice improvement would be if the suggestions api documentation [https://www.algolia.com/doc/rest-api/query-suggestions/](https://www.algolia.com/doc/rest-api/query-suggestions/) is updated with examples for the Go client (and other clients), like the excellent documentation for the search client :)

shortcuts:
> @shortcuts thanks for diving into it. Setting the region to EU seems to fix the issue indeed. Sorry for missing that. Closing this PR and the issue.

Glad it solved your issue!

> A nice improvement would be if the suggestions api documentation https://www.algolia.com/doc/rest-api/query-suggestions/ is updated with examples for the Go client (and other clients), like the excellent documentation for the search client :)

The new major version aims at making this clearer, hopefully the doc will help too!

> Still very strange that our "fix" in this PR made things work even with the wrong region configured by the way 🤯

yup that's weird 🤔 I guess this has to be something related with the 307 response, I'll ask the team in charge of that API if they redirect based on the app region

